PR #1, opened 2025-02-06
Initial commit of kmerseek search, index

olgabot:
This PR implements what was originally in the fairly messy notebooks:

- https://github.com/olgabot/botryllus-mhc/blob/main/notebooks/201_create_all_botryllus_protein_sigs.ipynb
- https://github.com/olgabot/botryllus-mhc/blob/main/notebooks/203_botryllus_run_gather_human_mouse.ipynb
- https://github.com/olgabot/botryllus-mhc/blob/main/notebooks/206a_e_values_per_kmer.ipynb

Goals of subcommands:

### `kmerseek index $database_fasta`

This creates everything necessary for a kmerseek search. This command wraps:

- Creation of k-mer sketches of database (uses [`sourmash scripts manysketch`](https://github.com/sourmash-bio/sourmash_plugin_branchwater/blob/main/doc/README.md#Running-manysketch) from [sourmash_plugins_branchwater](https://github.com/sourmash-bio/sourmash_plugin_branchwater/))
- Identification of hashval-to-kmer matches in database (uses modified [`sig2kmer.py`](https://github.com/olgabot/botryllus-mhc/blob/main/notebooks/sig2kmer.py) from [botryllus-mhc](https://github.com/olgabot/botryllus-mhc/) repo)
- Creation of RocksDB index in (uses [`sourmash scripts index`](https://github.com/sourmash-bio/sourmash_plugin_branchwater/blob/main/doc/README.md#Running-index) from [sourmash_plugins_branchwater](https://github.com/sourmash-bio/sourmash_plugin_branchwater/))

### `kmerseek search $query_fasta $database_index`

This function wraps everything necessary to perform the search. This command wraps:

- Creation of k-mer sketches of query (uses [`sourmash scripts manysketch`](https://github.com/sourmash-bio/sourmash_plugin_branchwater/blob/main/doc/README.md#Running-manysketch) from [sourmash_plugins_branchwater](https://github.com/sourmash-bio/sourmash_plugin_branchwater/))
- Identification of hashval-to-kmer matches in query (uses modified [`sig2kmer.py`](https://github.com/olgabot/botryllus-mhc/blob/main/notebooks/sig2kmer.py) from [botryllus-mhc](https://github.com/olgabot/botryllus-mhc/) repo)
- Search of query in database with [`multisearch`](https://github.com/sourmash-bio/sourmash_plugin_branchwater/blob/main/doc/README.md#Running-multisearch-and-pairwise) from [sourmash_plugins_branchwater](https://github.com/sourmash-bio/sourmash_plugin_branchwater/)
- TODO: Matching of protein domains from uniprotkb. It might look something like [this notebook](https://github.com/olgabot/botryllus-mhc/blob/main/notebooks/400_query_uniprot_rest_api.ipynb). However, the UniProt Rest API is rate-limited, and we might be better of pre-downloading the full [UniProtKB Reviewed (Swiss-Prot) release](https://www.uniprot.org/help/downloads) in XML format to parse. I think most use cases would search against UniProt, so I think this could be a valid approach


src/kmerseek/entity.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self, fasta, moltype, ksize, scaled):
+        # These are all filenames of where the data is stored
+        self.fasta = fasta
+        self.sketch_kws = dict(moltype=moltype, ksize=ksize, scaled=scaled)
+
+    @property
+    def sig(self):
+        """String of k-mer signature filename"""
+        if not hasattr(self, "_sig"):
+            self._sig = sketch(self.fasta, **self.sketch_kws)
```
mlovci, line 21: Of course, this works, but would it be preferable to use `@sig.setter` instead? [docs that make me think this](https://docs.python.org/3/library/functions.html#property)

src/kmerseek/search.py
```diff
@@ -56,7 +61,23 @@ def stitch_together_species_hp_kmers(
...
+    sourmash_plugin_branchwater.do_manysearch(
+        query_siglist,
+        target_siglist,
+        0,  # threshold=0 to show all matches, even with only 1 k-mer
+        ksize,
+        scaled,
+        moltype,
+        output,
+        False,  # Don't ignore abundance
+        False,  # Don't output ALL comparisons, only the onse with at least a match
```
mlovci, line 76: *onse -> ones

pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[project]
+name = "kmerseek"
+version = "0.1.0"
+description = "Efficient protein domain annotation search with reduced amino acid k-mers"
+authors = [{name = "Olga Botvinnik", email = "olga@seanome.org"}]
+dependencies = ["sourmash", "sourmash_plugin_branchwater", "pandas", "click"]
```
mlovci, line 6: Is `polars` requried?

src/kmerseek/index.py
```diff
@@ -0,0 +1,74 @@
...
+    return f"{sig}.rocksdb"
+
+
+def make_rocksdb_index(sig, moltype, ksize, scaled):
+    output = _make_rocksdb_filename(sig)
+    notify(f"indexing all sketches in '{sig}'")
+
+    siglist = _make_siglist_file(sig)
+
+    # Colros set to false because that's what the sourmash_plugin_branchwater code does
```
mlovci, line 41: *Colros -> colors

src/kmerseek/index.py
```diff
@@ -0,0 +1,74 @@
...
+    # in the same filesystem, "next door" in the same folder
+    internal_storage = False
+
+    status = sourmash_plugin_branchwater.do_index(
+        siglist,
+        ksize,
+        scaled,
+        moltype,
+        output,
+        colors,  # colors - currently must be false?
```
mlovci, line 54: why "?"

src/kmerseek/index.py
```diff
@@ -0,0 +1,74 @@
...
+        internal_storage,
+    )
+    if status == 0:
+        notify(f"...index is done! results in '{output}'")
+
+
+@click.command()
+@click.argument("fasta")
+@click.option("--moltype", default="hp")
```
mlovci, line 63: Should this be a ChoiceOption [click docs](https://click.palletsprojects.com/en/stable/options/#choice-options) so users may know which languages are supported?

src/kmerseek/index.py
```diff
@@ -0,0 +1,74 @@
...
+@click.option("--ksize", type=int, default=24)
+@click.option("--scaled", type=int, default=5)
+def index(fasta, moltype="hp", ksize=24, scaled=5):
```
mlovci, line 66: Defaults are set in the click decorators, should they be here too?

src/kmerseek/search.py
```diff
@@ -0,0 +1,169 @@
...
+from .sketch import sketch, make_sketch_kws
+from .index import KmerseekIndex, _make_siglist_file
+from .query import KmerseekQuery
+
+protein_moltypes = "protein", "dayhoff", "hp"
+
+
+def show_results(results_per_gene):
+    for gene, match in results_per_gene:
+        # This function currently doesn't do anything, just here for
```
mlovci, line 18: then remove it for now?

src/kmerseek/search.py
```diff
@@ -0,0 +1,169 @@
...
+    sourmash_plugin_branchwater.do_multisearch(
+        query_sig,
+        target_rocksdb,
+        0,  # threshold=0 to show all matches, even with only 1 k-mer
+        ksize,
+        scaled,
+        moltype,
+        False,  # No Average nucleotide identity (ANI) calculation
+        True,  # Yes, calculate probability of overlap between target and query
+        False,  # Don't output ALL comparisons, only the onse with at least a match
```
mlovci, line 91: *onse -> ones

src/kmerseek/sig2kmer.py
```diff
@@ -0,0 +1,217 @@
...
+        alphabet_func = sourmash._lowlevel.lib.sourmash_aa_to_dayhoff
+    else:
+        raise ValueError(f"Unknown moltype: {moltype}")
+
+    # Convert the entire sequence to bytes once
+    byte_sequence = sequence.encode("utf-8")
+
+    # Apply the alphabet function to each byte in the sequence and join the result
+    degenerate = b"".join(
+        alphabet_func(letter.to_bytes(1, "big")) for letter in byte_sequence
```
mlovci, line 33: Do you like `map` or `for`... i.e.:
b"".join(map(lambda s: alphabet_func(s.to_bytes(1, "big)), byte_sequence))

src/kmerseek/sig2kmer.py
```diff
@@ -0,0 +1,217 @@
...
+    return kmer_encoded, kmer_in_seq
+
+
+def revise_ksize(ksize, moltype, input_is_protein):
+    """If input is protein, then divide the ksize by three"""
+    if moltype == "DNA":
+        return ksize
+    elif input_is_protein:
+        # Ksize includes codons
+        return int(ksize / 3)
```
mlovci, line 77: is it important to check whether ksize mod 3 == 0?

src/kmerseek/sig2kmer.py
```diff
@@ -0,0 +1,217 @@
...
+        return ksize
+
+
+def get_kmers_for_hashvals(sequence, hashvals, ksize, moltype, input_is_protein):
+    "Return k-mers from 'sequence' that yield hashes in 'hashvals'."
+    # uppercase!
+    sequence = sequence.upper()
+
+    # Divide ksize by 3 if sequence is protein
+    # ksize = revise_ksize(ksize, moltype, input_is_protein)
```
mlovci, line 88: Remove for now?

src/kmerseek/sig2kmer.py
```diff
@@ -0,0 +1,217 @@
...
+    moltype,
+    input_is_protein,
+    hashes,
+    found_kmers,
+    m,
+    n,
+    n_seq,
+    seqout_fp,
+    kmerout_w,
+    watermark,
```
mlovci, line 115: Not clear what this arg is for... what is `watermark`?

src/kmerseek/sig2kmer.py
```diff
@@ -0,0 +1,217 @@
...
+                end="\r",
+            )
+            watermark += NOTIFY_EVERY_BP
+
+        # now do the hard work of finding the matching k-mers!
+        for kmer_encoded, kmer_in_seq, hashval, i in get_kmers_for_hashvals(
+            record.sequence, hashes, ksize, moltype, input_is_protein
+        ):
+            found_kmers += 1
+            # found_kmers.append([kmer_in_seq, kmer_encoded, hashval, record["name"]])
```
mlovci, line 137: remove for now?

tests/test_index.py
```diff
@@ -0,0 +1,19 @@
...
+    runner = CliRunner()
+    result = runner.invoke(cli, ["index", bcl2_first25])
+    assert result.exit_code == 0
+
+    # Make sure all the files got created
+    assert os.path.exists(f"{bcl2_first25}.manysketch.csv")
+    assert os.path.exists(f"{bcl2_first25}.hp.k24.scaled5.sig.zip")
+    assert os.path.exists(f"{bcl2_first25}.hp.k24.scaled5.sig.zip.kmers.csv")
+    assert os.path.exists(f"{bcl2_first25}.hp.k24.scaled5.sig.zip.siglist")
+    assert os.path.exists(f"{bcl2_first25}.hp.k24.scaled5.sig.zip.rocksdb")
```
mlovci, line 19: Is it important to check that the file has content, beyond just existing?

mlovci (commented):
Have not done a functional test, but have added some comments on readability and choices made. If there is energy to expand on documentation it may be useful, especially when it comes to sourmash options.

tests/test_index.py
olgabot: Probably.. this is a very basic smoke test for now. It assumes that the Sourmash functions worked properly, but yes, we should test content as well

src/kmerseek/index.py
olgabot: Probably not... Too much to maintain two different things

src/kmerseek/index.py
olgabot: comment was copied from Sourmash

src/kmerseek/index.py
olgabot: Probably! I didn't know about those, thanks!

pyproject.toml
olgabot: Yes, polars is required for dealing with the kmer csvs/parquet files

src/kmerseek/entity.py
olgabot: Hmm maybe? I asked Claude, Perplexity, and ChatGPT for what's the most Pythonic way to do this and they agreed about using `@property`. I think it's less code to deal with in using `@property` then separate getters and setters, but could be convinced otherwise!

Eventually would like to have checks if the files already exist, then don't do anything, which feels like it would be easier to write in the `@property` version

src/kmerseek/search.py
olgabot: Potentially, I'm still working on it. If I can get it working, then I'll keep it :)

src/kmerseek/sig2kmer.py
olgabot: Sorry, I don't understand the question

src/kmerseek/sig2kmer.py
```diff
@@ -0,0 +1,176 @@
...
+            moltype=moltype,  # or "dna", "dayhoff", "hp"
+            ksize=ksize,
+            scaled=scaled,
+            save_kmers=tmp_kmers.name,  # Use the temp csv path for save_kmers
+            save_sequences=tmp_fasta.name,  # Use the temp fasta path for save_sequences
+        )
+
+        # TODO: this "works" but calls what's normally a CLI as a Python method...
+        # would love for this to be just a regular Python method that you could
+        # feed sigs and fastas to
```
heuermh, line 164: This would require a code change upstream in sourmash?

src/kmerseek/sig2kmer.py
```diff
@@ -0,0 +1,176 @@
...
+
+    return kmers_with_encoding
+
+
+# TODO: Probably move this to `sourmash sig kmers` itself
+# i.e. to here: https://github.com/sourmash-bio/sourmash/blob/c209e7d39d80aa8eceed8d5a0c91568f96fded3f/src/sourmash/sig/__main__.py#L1087
+# https://github.com/seanome/kmerseek/issues/5
+def add_start_position_to_kmers_pl(kmers, fasta):
+    sequences = {}
+    with screed.open(fasta) as records:
```
heuermh, line 99: This is pulling all FASTA sequences into RAM?  May be worth converting FASTA --> Parquet and then doing a join across the two tables.  Is that possible in polars?

src/kmerseek/sig2kmer.py
```diff
@@ -0,0 +1,176 @@
...
+            self.dayhoff = True
+        elif moltype == "hp":
+            self.hp = True
+
+
+def add_encoding_to_kmers_pl(kmers, moltype):
+
+    # Apply encoding using a UDF (User Defined Function)
+    if moltype == "hp" or moltype == "dayhoff":
+        kmers_with_encoding = kmers.with_columns(
```
heuermh, line 80: Curious how UDF performance in polars compares to DuckDb and pyspark.  If you have this `kmers` dataframe without the encoding column saved to disk somewhere, I'd be interested in doing a little benchmarking

src/kmerseek/uniprot.py
```diff
@@ -0,0 +1,19 @@
+"""
+Parse JSON file from UniProt
+"""
+
+
+def get_domains(gene, start, end):
+    """For the gene provided, find overlapping regions
```
heuermh, line 7: Curious what `gene` is here, specifically?

src/kmerseek/search.py
```diff
@@ -0,0 +1,286 @@
...
+
+    # All these lengths should be the same -> if not, something went wrong
+    assert len(query) == len(alphabet)
+    assert len(alphabet) == len(match)
+
+    length = len(query)
+
+    match_start = df[start_match].min()
+    query_start = df[start_query].min()
+    match_end = match_start + length
```
heuermh, line 84: Would be nice to have the coordinate system explicitly documented somewhere for these, I assume 1-based fully closed intervals?

src/kmerseek/search.py
```diff
@@ -0,0 +1,286 @@
...
+            print(row["to_print"])
+
+
+@click.command()
+@click.argument("query_fasta")
+@click.argument("target_fasta")
+@click.option("--moltype", default="hp")
+@click.option("--ksize", default=24)
+@click.option("--scaled", default=5)
+@click.option("--output", default="kmerseek_search.csv")
```
heuermh, line 249: Is output to `stdout` an option?

heuermh (approved):
LGTM, left some questions

src/kmerseek/search.py
```diff
@@ -0,0 +1,286 @@
...
+import polars as pl
+
+from .uniprot import get_domains
+from .sketch import make_sketch_kws
+from .index import KmerseekIndex, _make_siglist_file
+from .query import KmerseekQuery
+from .logging import setup_logging, logger
+
+
+def show_results(results_per_gene):
```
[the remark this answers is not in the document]
olgabot: Yes, very good point. This `show_results` is also only good for looking at one gene at a time, and we need to output a CSV/parquet or similar of all matches and their k-mers

src/kmerseek/sig2kmer.py
olgabot: Yes, would require changing sourmash itself

src/kmerseek/search.py
```diff
@@ -0,0 +1,286 @@
...
+
+        self.per_gene_stitched_kmers = (
+            self.results_with_kmers.group_by("match_name")
+            .map_groups(stitch_kmers_in_query_match_pair, schema=output_schema)
+            .sort(["query_start", "query_end"])
+        ).collect()
+
+    def show_results_per_gene(self):
+        cols = self.per_gene_stitched_kmers.columns
+        for row_data in self.per_gene_stitched_kmers.iter_rows():
```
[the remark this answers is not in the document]
olgabot: Oh hmm interesting, that's a good idea. Would that be better than streaming? Or I guess the `sink_csv` output in polars *is* streaming, so we're good there!

src/kmerseek/sig2kmer.py
olgabot: Oh interesting, I hadn't thought about doing it that way. Thanks for the suggestion! I'll probably still use `screed.open` since it creates an iterator but can use `polars.LazyFrame` after that

src/kmerseek/uniprot.py
olgabot: Good point, technically this is the query protein sequence

src/kmerseek/search.py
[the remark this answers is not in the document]
heuermh: Keep in mind Parquet is column-based, not row based!  If the SQL query only needs to read one column and write another, it might be faster to not load entire rows off disk into RAM.

src/kmerseek/uniprot.py
olgabot: Oh actually it would be the UniProt ID. Let me change that.

src/kmerseek/search.py
olgabot: I'm thinking that `stdout` will show a pretty-print option and then the CSV will show the easily parse-able results, following how [`manysearch`](https://github.com/sourmash-bio/sourmash_plugin_branchwater/blob/97b1b2ba5aa36d001b762ab19cb21d757c43534d/src/manysearch.rs#L31) from sourmash branchwater operates:

```
kmerseek-dev)
 Tue 25 Feb - 11:14  ~/code/kmerseek   origin ☊ olgabot/separate-index-search-substeps ✔ 1☀ 
  sourmash scripts manysearch --help

== This is sourmash version 4.8.14. ==
== Please cite Irber et. al (2024), doi:10.21105/joss.06830. ==

usage:  manysearch [-h] [-q] [-d] -o OUTPUT [-t THRESHOLD] [-k KSIZE] [-s SCALED] [-m {DNA,protein,dayhoff,hp,skipm1n3,skipm2n3}] [-c CORES] [-P] [-N]
                   [--ignore-abundance] [-A]
                   query_paths against_paths

search many metagenomes for contained genomes

positional arguments:
  query_paths           input file of sketches
  against_paths         input file of sketches

options:
  -h, --help            show this help message and exit
  -q, --quiet           suppress non-error output
  -d, --debug           provide debugging output
  -o, --output OUTPUT   CSV output file for matches
  -t, --threshold THRESHOLD
                        containment threshold for reporting matches (default: 0.01)
  -k, --ksize KSIZE     k-mer size at which to select sketches
  -s, --scaled SCALED   scaled factor at which to do comparisons
  -m, --moltype {DNA,protein,dayhoff,hp,skipm1n3,skipm2n3}
                        molecule type: DNA, protein, dayhoff, hp, or skipmer (skipm1n3 or skipm2n3); default DNA
  -c, --cores CORES     number of cores to use (default is all available)
  -P, --pretty-print    display results after search finishes (default: True)
  -N, --no-pretty-print
                        do not display results (e.g. for large output)
  --ignore-abundance    do not do expensive abundance calculations
  -A, --output-all-comparisons
                        ignore threshold and output all comparisons; against a RocksDB database, this will only output comparisons with some overlap
```

I see the stdout and csv as providing two different kinds of views, where the `stdout` one is more human-parseable, while the `csv` is for filtering later. Does that make sense?

src/kmerseek/uniprot.py
heuermh: Which UniProt ID?  😄   `accession` for the entry, e.g. `P0C9F0`?

src/kmerseek/search.py
heuermh: `kmerseek search ... | cut -f 2,3 | grep -v foo | zstd --adapt -o filtered.txt.zst` is a kind of filtering later 😉

olgabot: Okay okay fine :) It now can output to csv. This comment was extremely helpful, because I realized this `kmerseek_search.csv` was lying, it was actually an intermediate file that was output from the `sourmash scripts multisearch/sourmash_plugin_branchwater.do_manysearch` command, not actually outputting the kmerseek results anywhere!

src/kmerseek/uniprot.py
olgabot: Haha thanks for the clarification! Yes, the primary accession like [`O43236`](https://www.uniprot.org/uniprotkb/O43236/entry). I'll make sure to clarify :)

src/kmerseek/sig2kmer.py
olgabot: The dataframe (with encoding, but easy enough to remove the column) is here: 

tests/testdata/fasta/bcl2_first25_uniprotkb_accession_O43236_OR_accession_2025_02_06.fasta.gz.hp.k24.scaled5.sig.TRUE.zip.kmers.pq

In this PR: https://github.com/seanome/kmerseek/pull/1/files#diff-cb93b0bf29da2052382df33788f3aa5d830724c60bce6d476a80423b9bdb165c

PR #14, opened 2025-03-13
olgabot/remove prefix environment

olgabot:
- **Remove 'prefix' from environment.yml**
- **Ignore .idea files**


heuermh (approved):
Works good for me on arm64 mac with
```bash
$ conda env create --file environment_mac_arm64.yml
```

Issue #9, opened 2025-02-14
Parallelize `sourmash sig kmers`!


olgabot:
`sourmash sig kmers` (wrapped in `kmerseek.sketch.get_kmers` called by `kmerseek.entity.KmerseekEntity`) is a single-threaded process that merges the provided signatures and iterates over the fasta file, which can be really slow.

For example, I've had this process running for 17 hours on the UniProt/SwissProt human data (only ~20k genes, while all of UniProt/SwissProt is ~500k!):

![Image](https://github.com/user-attachments/assets/98cd63ed-9fec-495d-ba39-ac1154f75603)

I'd love for this process to get parallelized somehow. Ideally we'd contribute upstream to [`sourmash`](https://github.com/sourmash-bio/sourmash) or [`sourmash_plugin_branchwater`](https://github.com/sourmash-bio/sourmash_plugin_branchwater/) itself.

olgabot:
Sourmash Sig Kmers code: https://github.com/sourmash-bio/sourmash/blob/latest/src/sourmash/sig/__main__.py (search for `def kmers`)

olgabot:
Update: still hasn't finished after 7 days

![Image](https://github.com/user-attachments/assets/996a0781-7cc8-4207-89f1-c18057e3f02f)

olgabot:
Update: Running for 12 days, 18 hours and still not done...

![Image](https://github.com/user-attachments/assets/217c047a-7139-45b6-83cc-684d0c77a74b)

olgabot:
Now we're at 12,000,874 residues (out of 207,621,817 residues in UniProt/SwissProt according to the [statistics](https://www.uniprot.org/uniprotkb/statistics))


![Image](https://github.com/user-attachments/assets/48dd7f09-e5cf-462d-9e0c-db3ff85c40a3)

olgabot:
From @bluegenes:

Iterator for `SeqToHashes` is the thing to change to get the encoded version and start position: https://github.com/sourmash-bio/sourmash/blob/db7d17911eb01f929535f50c5c41815914d530b4/src/core/src/signature.rs#L435

Could implement another iterator to return e.g. `Ok(kmer, hash)` instead of just `Ok(hash)`: https://github.com/sourmash-bio/sourmash/blob/db7d17911eb01f929535f50c5c41815914d530b4/src/core/src/signature.rs#L489

bluegenes:
> From [@bluegenes](https://github.com/bluegenes):
> 
[...]

PR #17, opened 2025-04-01
Add skipping of kmer extraction

olgabot:
Addresses https://github.com/seanome/kmerseek/issues/16, ping @heuermh 

In this PR, we are defaulting to NOT extracting the k-mer sequences from each of the inputs until #9 is finished, because it bottlenecks getting at least initial results of matches per sequence, even if we don't get to see exactly which k-mers they are yet. Maybe that can be a post-processing step.


src/kmerseek/index.py
```diff
@@ -62,13 +62,18 @@ def make_rocksdb_index(sig, moltype, ksize, scaled):
 @click.option("--moltype", default="hp")
 @click.option("--ksize", type=int, default=24)
 @click.option("--scaled", type=int, default=5)
+@click.option("--do-kmer-extraction", is_flag=True, default=False)
```
heuermh, line 65: If the default is to not run kmer extraction, I would suggest using `--extract-kmers` as the option name.

If the default is to run kmer extraction, I would suggest using `--skip-extract-kmers` as the option name.

src/kmerseek/search.py
```diff
@@ -263,11 +265,24 @@ def write_to_file(self, filename):
...
             ]
         )
-        #
-        # import pdb
-        #
-        # pdb.set_trace()
 
+    def process_results(self, filename):
+        search_results = self.read_results()
+        if self.do_kmer_extraction:
```
heuermh, line 271: Passing boolean flags down leads to something like this, where one method has two different behaviors.  Which might be fine in the short term, but what happens when there is a third or fourth?

I wonder if any of the following refactorings might look better?

* Create a new top level command `extract-kmers`
* Create a new `KmerseekExtractKmers` at the same level as `KmerseekIndex`/`KmerseekSearch`
* Create two [specializations](https://en.wikipedia.org/wiki/UML_class_diagram#Generalization/Inheritance) each for `KmerseekIndex`/`KmerseekSearch`, one extracting kmers and one that does not
* Extract the `if self.do_kmer_extraction: ...` and `else:` parts to separate [strategy](https://en.wikipedia.org/wiki/Strategy_pattern) classes, which can be passed in and delegated to

heuermh (commented):
Left a couple suggestions to consider

src/kmerseek/index.py
olgabot: Thank you for the suggestion! Changed

src/kmerseek/search.py
```diff
@@ -173,25 +178,23 @@ def __init__(
...
+        if filename is None:
+            sys.stdout.write(df.write_csv())
+        else:
+            df.write_csv(filename)
+
+    def process_results(self, filename):
+        raise NotImplementedError("Subclasses must implement this method.")
+
+
+class KmerseekResultsWithKmerExtraction(KmerseekResultsBase):
```
olgabot, line 193: @heuermh is this the kind of thing you meant by the specializations and strategy classes? Did I interpret that corectly?

src/kmerseek/extract_kmers.py
```diff
@@ -0,0 +1,66 @@
...
+        """Polars LazyFrame of all the kmers."""
+        if self._kmers_lazyframe is None:
+            if self.kmers_pq:
+                self._kmers_lazyframe = pl.scan_parquet(self.kmers_pq)
+            else:
+                logger.warning("K-mer parquet file not generated.")
+                return None
+        return self._kmers_lazyframe
+
+    def run_extraction(self):
```
heuermh, line 63: Maybe `run_extraction` &rarr; `extract_kmers`?

src/kmerseek/search.py
heuermh: Yep!  The long class names are kinda sucky but they are clear.  What do you think?

heuermh (approved):

src/kmerseek/search.py
olgabot: Awesome! Wanted to make sure I implemented it right. Long class names are okay, as long as they're not weirdly repetitive like `KmerseekResultsFactoryParserFactoryReader` 😆

src/kmerseek/extract_kmers.py
olgabot:
```suggestion
    def extract_kmers(self):
```

Yes, much clearer!

PR #37, opened 2026-08-07
Add --remove-low-complexity to drop homopolymer k-mers at index time

olgabot:
Me: The most common hydrophobic-polar k-mers are overwhelmingly {h}_n and {p}_n homopolymer runs, which skew the results because of spurious matches to repetitive sequences. This adds the option to remove low complexity k-mers when building the index, and also takes into account the low complexity k-mers at search time, so the additional low complexity k-mers aren't adding to the denominator in the Jaccard calculation.

I'd like this to be included in v0.4.0 https://github.com/seanome/kmerseek/pull/34 for future testing

---
## What

Adds an opt-in `--remove-low-complexity` flag to `kmerseek index`. Low-complexity k-mers — homopolymer runs like a poly-glutamate tract (`EEEEE`) or, under a reduced alphabet, an all-hydrophobic window (`hhhhh`) — are abundant and carry little discriminative signal.

**Off by default.** Existing indexes and workflows are unaffected.

```bash
kmerseek index -i proteome.fasta --ksize 10 --encoding hp --remove-low-complexity
```

## Two independent checks

Per k-mer, when enabled:

1. **Raw amino-acid** window — any encoding. Catches `EEEEE`.
2. **HP-encoded** window — `hp`-family encodings only. Catches windows that aren't raw homopolymers but still collapse to one symbol: `LIVMA` is five different residues that all encode to `h`.

Only *exact* homopolymers are removed. A near-homopolymer like `hhhhhhhhhp` is kept — see follow-up below.

## The setting is stored in the index, and search reuses it

This is the part that matters most for correctness. Containment is `intersection / query_size`. If the index removed these k-mers but queries kept them, those k-mers would match nothing while still inflating the denominator — deflating scores for exactly the queries that contain low-complexity regions.

So the flag is persisted and read back at search time. You don't repeat it when searching:

```
Remove low-complexity k-mers: true (from index)
```

Verified on the bcl2 test set: containment shifts on **all 25 shared query/target pairs**, in both directions — the denominator shrinks while some intersecting k-mers also drop.

### Persistence, and index versioning

`remove_low_complexity` is a real field on `ProteomeIndexMetadata`, and `save_state` stamps the writing kmerseek version (`CARGO_PKG_VERSION`) into a `kmerseek_version` key.

bincode is not self-describing, so that key doubles as the metadata layout marker: stamped indexes deserialize with the current struct, unstamped ones go through `LegacyProteomeIndexMetadata` and default the flag to `false` — correct, since they kept every k-mer. Old indexes still open; the field is first-class going forward. `SCHEMA_VERSION` goes 1 → 2 to reflect the layout change, and `read_kmerseek_version()` is public so you can check what built an index.

### Counting costs nothing extra

The removal tallies accumulate into two `AtomicUsize` totals on `ProteomeIndex` as each signature is built, so `low_complexity_counts()` is an O(1) load with no second pass over the signature map. Atomics because `create_protein_signature` takes `&self` and `process_fasta` drives it from a `par_iter`.

## Filenames don't collide

The point of the flag is A/B comparison, so auto-generated names gain a `.nolowcomplexity` segment. Otherwise indexing the same FASTA both ways would silently clobber the first database.

```
proteome.fasta.hp.k10.scaled1.kmerseek.rocksdb                  # default
proteome.fasta.hp.k10.scaled1.nolowcomplexity.kmerseek.rocksdb  # --remove-low-complexity
```

## Indexing reports what it removed

So you can tell whether the flag mattered without rebuilding and diffing:

```
Removed 75 of 9063 k-mer windows as low-complexity (0.83%)
```

## Drive-by cleanup

Removes `ProteomeIndex::process_kmers` — a second, unfiltered k-mer walk that nothing but tests still called (`create_protein_signature` stopped using it). It turned out to be the sole consumer of the `encoding_fn` field and four imports, so those came out too. Moltype validation is preserved at each site by the existing `get_hash_function_from_moltype` call. Its tests already exercised `add_protein` directly and are renamed to match.

## Naming

`remove` rather than `filter` throughout, since "filter" doesn't say which direction — filter *in* or filter *out*?

## Testing

**142 tests pass**, clean build with zero warnings, `cargo fmt --check` clean.

Seven tests cover this feature specifically:

- `is_homopolymer_kmer` detects runs / rejects mixed k-mers
- Toggle on/off at the `ProteinSketch` level and through `ProteomeIndex`
- Raw amino-acid removal for plain `protein` moltype, using a genuine 11-residue poly-E tract from human FKBP8 (UniProt Q14318, residues 26–55) already in the repo's test FASTA — real biology, not an invented motif
- Persistence round-trip through `save_state` → `open_for_search`
- A forged pre-versioning index still loads through the legacy metadata layout, with the flag defaulting to `false`
- `save_state` stamps the kmerseek version
- Custom `hp_*` alphabets, which take a different encode-and-check path from the built-in `hp`
- Index-level count aggregation across multiple signatures, and the builder's `remove_low_complexity`
- CLI round-trip: index with the flag, then search without it and assert the setting is recovered
- CLI filename collision

## Follow-up, not in this PR

Broadening beyond exact homopolymers. A k-mer frequency histogram on real data shows the top offenders are 9-of-10 homopolymers that this PR's strict check lets through:

```
hhhhhhhhhp: 653
hppppppppp: 651
phhhhhhhhh: 645
```

Catching those needs a fuzzier threshold (max run length, or "at most N minority residues") — a separate design decision.

### DIFF
--- a/src/rust/index.rs
+++ b/src/rust/index.rs
@@ -12,7 +12,6 @@ use flate2::write::GzEncoder;
 use flate2::Compression;
 use rocksdb::{Options, DB};
 use serde::{Deserialize, Serialize};
-use sourmash::_hash_murmur;
 use sourmash::collection::Collection;
 use sourmash::manifest::Manifest;
 use sourmash::signature::SigsTrait;
@@ -20,11 +19,8 @@ use sourmash::sketch::minhash::KmerMinHash;
 use sourmash::storage::{FSStorage, InnerStorage};
 
 use crate::aminoacid::AminoAcidAmbiguity;
-use crate::encoding::{
-    encode_with_fn, get_encoding_fn_from_moltype, get_hash_function_from_moltype,
-};
+use crate::encoding::get_hash_function_from_moltype;
 use crate::errors::{IndexError, IndexResult};
-use crate::hp_alphabets::HpAlphabet;
 use crate::signature::{SignatureAccess, SEED};
 use crate::sketch::{ProteinSketch, ProteinSketchStore};
 
@@ -97,9 +93,6 @@ pub struct ProteomeIndex {
     // Amino acid ambiguity handler
     aa_ambiguity: Arc<AminoAcidAmbiguity>,
 
-    // Protein encoding function
-    encoding_fn: fn(u8) -> u8,
-
     // Statistics for k-mer frequencies and IDF
     // Not currently used, but will be used in the future
     #[allow(dead_code)]
@@ -123,6 +116,12 @@ pub struct ProteomeIndex {
 
     // Configuration for raw sequence storage
     store_raw_sequences: bool,
+
+    // Whether to drop low-complexity (homopolymer) k-mers when building protein
+    // signatures: raw amino-acid runs for any moltype, plus all-h/all-p runs for
+    // HP-family moltypes. Defaults to false. Persisted separately from the
+    // metadata blob; see save_state and read_remove_low_complexity.
+    remove_low_complexity: bool,
 }
 
 impl Drop for ProteomeIndex {
@@ -217,9 +216,6 @@ impl ProteomeIndex {
         let hash_function = get_hash_function_from_moltype(moltype)
             .map_err(|e| IndexError::SourmashError(e.to_string()))?;
 
-        let encoding_fn = get_encoding_fn_from_moltype(moltype)
-            .map_err(|e| IndexError::SourmashError(e.to_string()))?;
-
         let minhash_ksize = ksize * 3;
         // Create the minhash sketch
         let minhash = KmerMinHash::new(
@@ -242,14 +238,51 @@ impl ProteomeIndex {
             signatures: DashMap::new(),
             combined_minhash: Arc::new(Mutex::new(minhash)),
             aa_ambiguity: Arc::new(AminoAcidAmbiguity::new()),
-            encoding_fn,
             moltype: moltype.to_string(),
             ksize,
             minhash_ksize,
             scaled,
             stats: ProteomeIndexKmerStats { idf: HashMap::new(), frequency: HashMap::new() },
             store_raw_sequences,
+            remove_low_complexity: false,
         })
+    }
+
+    /// Enable or disable dropping low-complexity (homopolymer) k-mers when
+    /// building protein signatures. Defaults to `false`.
+    pub fn set_remove_low_complexity(&mut self, remove_low_complexity: bool) {
+        self.remove_low_complexity = remove_low_complexity;
+    }
+
+    /// Whether this index drops low-complexity (homopolymer) k-mers.
+    ///
+    /// Search reads this to build query sketches the same way the targets were
+    /// built; see `save_state` for why a mismatch skews containment.
+    pub fn remove_low_complexity(&self) -> bool {
+        self.remove_low_complexity
+    }
+
+    /// Total k-mer windows examined across all in-memory signatures, and how many
+    /// were removed as low-complexity. Returns `(0, 0)` when removal is off.
+    ///
+    /// Only meaningful right after indexing: the counts live on the in-memory
+    /// sketches and are not persisted.
+    pub fn low_complexity_counts(&self) -> (usize, usize) {
+        self.signatures.iter().fold((0, 0), |(examined, skipped), entry| {
+            let (e, s) = entry.value().low_complexity_counts();
+            (examined + e, skipped + s)
+        })
+    }
+
+    /// Read the persisted low-complexity removal setting from an open database.
+    ///
+    /// Indexes built before this flag existed have no such key, and were by
+    /// definition kept every k-mer, so a missing key reads as `false`.
+    fn read_remove_low_complexity(db: &DB) -> IndexResult<bool> {
+        match db.get(b"remove_low_complexity")? {
+            Some(data) => Ok(bincode::deserialize(&data)?),
+            None => Ok(false),
+        }
     }
 
     /// Get a reference to the signatures map (for testing)
@@ -743,8 +776,17 @@ impl ProteomeIndex {
         // deserializing the full metadata (and without breaking old bincode layouts).
         let serialized_version = bincode::serialize(&SCHEMA_VERSION)?;
         self.db.put(b"schema_version", serialized_version)?;
+
+        // Store the low-complexity removal setting as its own key, for the same reason:
+        // adding a field to ProteomeIndexMetadata would break bincode reads of indexes
+        // built before this flag existed. Absent key means "kept everything" (see
+        // read_remove_low_complexity), which is exactly right for those older indexes.
+        // WHY persist at all: search must build query sketches the same way, or
+        // retained low-complexity query k-mers inflate the containment denominator.
+        let serialized_filter = bincode::serialize(&self.remove_low_complexity)?;
+        self.db.put(b"remove_low_complexity", serialized_filter)?;
         eprintln!(
-            "[save] Metadata + schema_version written in {:.1}s total",
+            "[save] Metadata + schema_version + remove_low_complexity written in {:.1}s total",
             t3.elapsed().as_secs_f32()
         );
 
@@ -951,8 +993,6 @@ impl ProteomeIndex {
 
             let _hash_function = get_hash_function_from_moltype(&metadata.moltype)
                 .map_err(|e| IndexError::SourmashError(e.to_string()))?;
-            let encoding_fn = get_encoding_fn_from_moltype(&metadata.moltype)
-                .map_err(|e| IndexError::SourmashError(e.to_string()))?;
 
             // Reconstruct the combined minhash from raw data
             let hash_function = get_hash_function_from_moltype(&metadata.moltype)
@@ -998,6 +1038,7 @@ impl ProteomeIndex {
             let moltype = &metadata.moltype;
             let ksize = metadata.ksize;
             let scaled = metadata.scaled;
+            let remove_low_complexity = Self::read_remove_low_complexity(&db)?;
 
             let signatures: DashMap<String, ProteinSketch> = DashMap::new();
             raw_chunks.par_iter().try_for_each(|raw_data| -> IndexResult<()> {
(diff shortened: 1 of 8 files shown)


src/rust/index.rs
```diff
@@ -238,16 +234,53 @@ impl ProteomeIndex {
...
+    /// building protein signatures. Defaults to `false`.
+    pub fn set_remove_low_complexity(&mut self, remove_low_complexity: bool) {
+        self.remove_low_complexity = remove_low_complexity;
+    }
+
+    /// Whether this index drops low-complexity (homopolymer) k-mers.
+    ///
+    /// Search reads this to build query sketches the same way the targets were
+    /// built; see `save_state` for why a mismatch skews containment.
+    pub fn remove_low_complexity(&self) -> bool {
```
olgabot, line 337: Is this necessary? Can't we just make the property public?

src/rust/index.rs
```diff
@@ -238,16 +234,53 @@ impl ProteomeIndex {
...
+        self.remove_low_complexity
+    }
+
+    /// Total k-mer windows examined across all in-memory signatures, and how many
+    /// were removed as low-complexity. Returns `(0, 0)` when removal is off.
+    ///
+    /// Only meaningful right after indexing: the counts live on the in-memory
+    /// sketches and are not persisted.
+    pub fn low_complexity_counts(&self) -> (usize, usize) {
+        self.signatures.iter().fold((0, 0), |(examined, skipped), entry| {
```
olgabot: always use par_iter

olgabot: Also is this necessary to re-iterate over all the signatures after building them? Couldn't we return this info as they're built?

src/rust/index.rs
```diff
@@ -461,8 +494,17 @@ impl ProteomeIndex {
         // deserializing the full metadata (and without breaking old bincode layouts).
         let serialized_version = bincode::serialize(&SCHEMA_VERSION)?;
         self.db.put(b"schema_version", serialized_version)?;
+
+        // Store the low-complexity removal setting as its own key, for the same reason:
+        // adding a field to ProteomeIndexMetadata would break bincode reads of indexes
+        // built before this flag existed. Absent key means "kept everything" (see
```
olgabot: we don't really need to care about previous indices.. I mean we should be mildly backwards compatible but I'd rather have this field moving forward. Can we Add the kmerseek version to the index and read the field that way?

olgabot (commented):
Need to make sure code is efficient, not re-iterating over all the same data. Also questions about public/private properties

src/rust/index.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Pushed back on this one — kept the getter, happy to change if you still want the field public.

Two reasons. Every sibling field on `ProteomeIndex` is private with a getter (`ksize()`, `scaled()`, `moltype()`, `store_raw_sequences()`), so making this one `pub` would be the odd one out. More importantly, it is now read from persisted metadata at load time and must stay in sync with what is on disk — a `pub` field lets a caller flip it after loading and silently desync the in-memory flag from the index it came from, which is exactly the containment-skew bug the persistence is there to prevent. The `set_` method is deliberately the only mutation point, for use before indexing.

src/rust/index.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Good catch — fixed in 3feaa4c, and this made the `par_iter` suggestion moot, so I did this instead.

`ProteomeIndex` now carries two `AtomicUsize` totals that `create_protein_signature` folds into while it already has the sketch in hand. `low_complexity_counts()` is an O(1) atomic load, so there is no second pass to parallelize. Atomics rather than `&mut` because `create_protein_signature` takes `&self` and `process_fasta` already drives it from a `par_iter`.

olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Superseded — see the reply on your other comment here. Rather than parallelizing the extra pass, 3feaa4c removes it: the counts accumulate during the build, so the read is now an O(1) atomic load with no iteration to parallelize.

src/rust/index.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Done — though the first attempt (3feaa4c) got this partly wrong, and 1d3f639 corrects it. Recording both so the thread is not misleading.

**Where it landed:** `remove_low_complexity` is a real field on `ProteomeIndexMetadata`, and `save_state` stamps `CARGO_PKG_VERSION` into a `kmerseek_version` key.

**What I got wrong first:** I used the *presence* of that version key to decide the metadata layout. That does not scale — 0.4 and 0.5 both carry the key, so it could never distinguish a future layout change. Layout is now keyed off `schema_version` (bumped 1 → 2), which is monotonic and exists for exactly this. `kmerseek_version` is purely provenance.

And it is now actually read back rather than write-only: `ProteomeIndex` stores it, `kmerseek_version()` exposes it, and search reports `Index built by kmerseek: 0.3.1`. `read_kmerseek_version` is no longer `pub` — it took `&rocksdb::DB`, leaking the storage dependency into the public API for one test that, being a child module, could already see it.

Old indexes still open: anything below schema 2 reads through `LegacyProteomeIndexMetadata` with the flag defaulting to `false`, correct since they kept every k-mer.

Covered by `test_unversioned_index_reads_through_legacy_metadata_layout` (forges both a schema-1 index and one with no `schema_version` key at all), `test_save_state_stamps_kmerseek_version`, and `test_load_round_trips_flag_version_and_signatures`.

src/rust/main.rs
```diff
@@ -148,7 +156,15 @@ fn main() -> IndexResult<()> {
     eprintln!("kmerseek {}", env!("CARGO_PKG_VERSION"));
 
     match cli.command {
-        Commands::Index { input, output, ksize, encoding, shuffled_seed, progress_interval } => {
```
olgabot: Let's also add --remove-low-complexity to the search tool

src/rust/main.rs
```diff
@@ -445,6 +518,7 @@ fn main() -> IndexResult<()> {
 
                     let mut query_sig =
                         ProteinSketch::new(name, final_ksize, final_scaled, final_encoding.into())?;
+                    query_sig.set_remove_low_complexity(remove_low_complexity);
```
olgabot, line 521: how much time does removing low complexity add to index and search?

src/rust/main.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Done in f3c1db6.

`search` now takes `--remove-low-complexity`, with an optional value so `--remove-low-complexity false` works too. Omit it and search follows whatever the index was built with, which is the usual case and needs no flag.

Output states the index property and the search behavior as two separate facts, rather than one line that conflated them:

```
Index: low-complexity k-mers were REMOVED when it was built
This search: low-complexity k-mers are REMOVED from query sketches (matching the index)
```

Overriding is allowed but warns, since containment is `intersection / query_size` and k-mers dropped on only one side still count toward the denominator:

```
This search: low-complexity k-mers are REMOVED from query sketches (--remove-low-complexity)
WARNING: this disagrees with the index. ...
```

Results also carry a `remove_low_complexity` column next to `ksize`/`scaled`/`moltype`, so a CSV is self-describing without the command line that produced it.

I used a column rather than a `#` comment line deliberately: `scripts/visualize_hits.py` reads results with `pl.scan_csv` and no `comment_prefix`, so a comment line would be parsed as the header and break it. pandas and R default the same way. Verified polars still parses the new output, as a Boolean.

src/rust/main.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Measured it. **Short answer: no measurable cost on either side — both deltas are smaller than the run-to-run noise.**

### Method

Release build (debug numbers would be meaningless), `k=10`, `hp`.

Input is the repo's `uniprotkb_..._Uncharacterized` FASTA (2841 real proteins, 674k residues) expanded 10x to 6.74M residues. Each copy is a rotation of the original sequence, so amino acid composition stays exactly real — which matters when benchmarking a composition-sensitive filter — while every record stays distinct, since identical sequences would collapse to one md5 and silently shrink the workload.

### Indexing, 6.74M residues, n=5

| | mean | median | min–max | sd |
|---|---|---|---|---|
| keep | 2.454s | 2.424s | 2.410–2.559 | 0.062 |
| remove | 2.488s | 2.488s | 2.436–2.544 | 0.038 |

Delta **+0.034s (+1.4%)**. The spread within the `keep` condition alone is 0.149s — over 4x the delta.

### Search, 50 queries vs 2841 targets, n=4

Same index for both runs, toggling only query-side removal via the new override, so target content is held constant and this isolates the query-sketching cost.

| | mean | median | min–max | sd |
|---|---|---|---|---|
| keep | 11.566s | 11.535s | 11.469–11.723 | 0.113 |
| remove | 11.650s | 11.515s | 11.292–12.278 | 0.438 |

Delta **+0.084s (+0.7%)** by mean, but **−0.2% by median** — the means differ only because of one 12.278s outlier. The two are not distinguishable.

### Why it is this cheap

`is_homopolymer_kmer` short-circuits on the first byte that differs. In real protein sequence almost every window fails on the second residue, so the common case is one or two byte comparisons — negligible next to murmur hashing and RocksDB writes.

On the index side it is not even purely additive: with removal on, our own window loop replaces sourmash's internal windowing rather than running in addition to it.

It also slightly *reduces* downstream work, since the removed k-mers are never hashed, stored, or searched. On the bcl2 test set at k=10 that was 75 of 9063 windows (0.83%).

### Caveat

Both deltas are within noise at these sample sizes, so treat these as upper bounds (low single-digit percent) rather than point estimates. Worth re-checking at proteome scale, where I would expect the picture to hold or slightly favor removal, since the index gets smaller.

PR #38, opened 2026-08-08
Region-scoped scoring for matched regions

olgabot:
## Why

`search.rs` only scored the whole protein. `MatchedRegion` carried `query_start`/`query_end` as positional metadata, but nothing used them to shrink the background count. A tight sub-protein domain match was diluted by the whole protein's k-mer count and could not be thresholded on its own.

## What's added

Per matched region, alongside every existing column:

| column | meaning |
|---|---|
| `region_n_shared_kmers` | `region_length - ksize + 1` |
| `region_expected_shared_kmers` | expected shared k-mers by chance, computed only from k-mer positions inside the region, not the whole protein |
| `region_poisson_score` | -log10 of the same Poisson survival-function probability as the whole-protein p-value, using the region-scoped expectation above. Bigger is more surprising. See "This is a ranking score, not a calibrated p-value" below |
| `region_tail_probability` | the raw probability `region_poisson_score` was computed from, before the -log10 transform: `10^(-region_poisson_score)`. Reported so downstream code that needs a real probability (e.g. Benjamini-Hochberg correction) doesn't have to invert the transform. Same caveats as `region_poisson_score` -- not a calibrated p-value either |
| `region_enrichment` | observed shared k-mers divided by `region_expected_shared_kmers` |

Four more columns record how many other things were tested alongside this one, without folding that count into the score:

- `region_search_space`: how many places in the query a region like this one could have started (`query_length - ksize + 1`)
- `db_n_targets`: how many target sequences were searched
- `db_n_kmers`: total k-mer occurrences across the database
- `run_n_queries`: how many queries were in this search run

These four are kept as separate numbers instead of being multiplied into `region_poisson_score`. The reasoning is in the next section.

### The math

For a query k-mer with hash `h`, let `freq_target(h)` be the number of target signatures containing that hash, and `N` be `db_n_targets`. For a region spanning query positions `[start, end)`, and `ksize` the k-mer size:

```
lambda_region = sum over every query k-mer whose start position p satisfies start <= p < end - ksize + 1
                    of  freq_target(h) / N

n_shared = region_length - ksize + 1                 (region_n_shared_kmers)

pvalue = P(X >= n_shared), where X ~ Poisson(lambda_region)
       = 1 - PoissonCDF(n_shared - 1, lambda_region)

region_poisson_score = -log10(pvalue)                 (bigger is more surprising; 0.0 at pvalue = 1.0)

region_enrichment = n_shared / lambda_region
```

There is no division by the region's length anywhere in `lambda_region`: each term in the sum is already a per-k-mer database frequency (`freq_target(h) / N`), and the sum has one term per k-mer position inside the region, not an average over the region's span. This is the same formula already used for the whole-protein `query_poisson_pvalue` (which is reported as a plain p-value, not log-transformed), with the sum restricted to k-mers inside the region instead of the whole query.

`region_poisson_score` is reported as -log10(p) rather than p directly so bigger consistently means "more evidence" across every column in this PR (`region_poisson_score`, `region_enrichment`, `query_enrichment` are all bigger-is-better; only `query_poisson_pvalue`, kept as a plain p-value, is smaller-is-better). A score of 0.0 means no evidence at all (p = 1.0); a score of about 1.3 is p = 0.05; a score of about 3.16 is p = 0.0007, the real BCL2/CED9 landmark below.

### This is a ranking score, not a calibrated p-value

`region_poisson_score` is named `_score`, not `_pvalue`, on purpose. It should be used to rank and prioritize candidate regions against each other, not to make a significance claim about any single region. Two problems keep the underlying probability from being a real p-value, and the -log10 transform changes neither of them.

First, `n_shared` is not an independent observation. It equals `region_length - ksize + 1`, arithmetic on the region's own length. The region's length is what `find_matched_regions` picked by keeping the longest gapless run of shared k-mers between the two sequences. The Poisson test is applied to the same quantity that defined the region's boundaries: it answers "how surprising is a window with this many matches," for a window that was chosen because it has this many matches. A p-value formula assumes the window was fixed before looking at the data. Here it was not.

Second, the k-mers being counted overlap. Two k-mers whose start positions are one residue apart share `ksize - 1` residues. A run of 5 overlapping k-mers spanning 19 residues is not 5 independent successes; it is closer to one match event observed 5 times over. The Poisson model assumes independent trials. Counting overlapping k-mers as independent understates how likely a run this long is to appear by chance, on top of the window-selection problem above.

### The search-space correction, and what it does and doesn't tell you

The standard fix for "I tested many candidate windows and reported the best one" is a multiple-testing correction: multiply the underlying probability by roughly how many candidates were tried, so a good-looking result found by chance among many candidates does not look as impressive as the same result found from a single planned test. `region_search_space` (how many places a region could have started) and `db_n_targets` (how many targets were searched) give that count here. Converting the score back to a probability first:

```
region_evalue = 10^(-region_poisson_score) * region_search_space * db_n_targets
```

Measured on BCL2 vs CED9 at k=15 (`region_poisson_score` = 3.16, i.e. p = 0.00069):

```
region_evalue = 0.00069 × 266 × 25
              = 4.6
```

Treat 4.6 as an order-of-magnitude estimate, not a precise corrected p-value. This multiplication corrects for having picked the best-looking window out of many candidates. It does not correct for the two problems above, so the input probability (0.00069) was already too small before this multiplication. The raw probability overstates this hit by more than the roughly 6600× the arithmetic above suggests. The honest reading of 4.6 is "this number needs real calibration before it supports any significance claim," not "a match this good happens 4.6 times by chance."

This PR reports the three raw numbers, `region_poisson_score`, `region_search_space`, `db_n_targets`, separately instead of pre-multiplying them, so anyone using this data can apply a correction, or replace the approach with a better one.

### What a calibrated version would need

A properly calibrated statistic here would model the length of the longest gapless run of shared k-mers directly, using an extreme-value distribution (the same family of model behind BLAST's E-values, via Karlin-Altschul statistics), account for the k-mer overlap (see Blanca et al. 2022 for the correlation structure this introduces), and be checked empirically: run the same search against a composition-matched, shuffled decoy database (not reversed; see Glidden-Handgis & Wheeler 2024, https://doi.org/10.1093/bioadv/vbae052) and measure how often a region this good turns up by chance, following the approach in Eddy 2008 (https://doi.org/10.1371/journal.pcbi.1000069).

None of that is implemented in this PR. This PR adds the region-scoped counts and the search-space components that calibration would need, and documents, here and in the code, why the current numbers rank candidates but do not support a significance claim. Whether to invest in full calibration, or to present the region score explicitly as a ranking/triage statistic rather than an inferential one, is an open decision this PR does not make.

### Two more caveats

`region_expected_shared_kmers` uses background k-mer frequencies computed over the whole database, not the region's own local composition. A region sitting in an unusual stretch of the protein, for example an unusually hydrophobic stretch, looks more enriched than it should, because it is compared against the database average rather than against similar local sequence. This is documented on `MatchedRegion::expected_shared_kmers` in the code.

`run_n_queries` is reported but never used in any calculation. Folding it in would make a match's reported significance depend on what else happened to be in the same search batch, so rerunning one query by itself would change the number. A test now checks that a query's score and `region_search_space` come out identical whether it runs alone or as part of a batch. `run_n_queries` would also be the wrong correction to reach for on its own: the number of targets searched and the number of candidate windows per target are both far larger than the number of queries in a typical run, so correcting by query count alone would understate the real search space. `region_search_space` and `db_n_targets`, not `run_n_queries`, are the dominant terms, which is why they are the two multiplied into the E-value above.

## Renames

| before | after |
|---|---|
| `poisson_pvalue` | `query_poisson_pvalue` |
| `expected_shared_kmers` | `query_expected_shared_kmers` |
| `enrichment` | `query_enrichment` |
| `query_start` / `query_end` / `query_subseq` | `region_start` / `region_end` / `region_subseq` |
| `MatchedRegion::poisson_pvalue` / `region_poisson_pvalue` (CSV) | `MatchedRegion::poisson_score` / `region_poisson_score`, now -log10(p) instead of p directly |
| `SearchFilters::max_region_pvalue` / `--max-region-pvalue` | `SearchFilters::min_region_score` / `--min-region-score`, and the comparison flipped from "keep if below" to "keep if above" |

The region-scoped number went through two changes during review, for the same underlying reason: it is not a calibrated p-value (see above), so neither its name nor its scale should look like one. First it was renamed from `_pvalue` to `_score`. Then it was put on a -log10 scale so that, like every other column in this PR, bigger consistently means more evidence: a plain probability is the one place in this schema where smaller is better, which invites the "threshold it at 0.05 by reflex" mistake this section is trying to prevent. `query_poisson_pvalue` keeps its name and its plain-probability scale; the whole-query scope does not have the region scope's window-selection problem, so it is a more standard (if still imperfect, given the same k-mer-overlap caveat) p-value.

`MatchedRegion`'s own fields keep no prefix, since the struct already means one region, matching the existing `length` field. `SearchResultCsv` adds a `region_` prefix when flattening into a row, the same way it already did for `region_length`.

## CLI

Two new flags: `--max-query-pvalue` (default 0.05, smaller passes) and `--min-region-score` (default ~1.301, i.e. p=0.05 on the -log10 scale, bigger passes). A match is kept if either one passes. `--min-region-score` is a cutoff on a ranking score, not a statistically calibrated significance threshold; see the statistics section above.

Requiring both to pass would reintroduce the dilution problem this PR fixes. Searching CED9 against the BCL2 family at default settings returned 0 matches before this change, even though CED9 carries a region matching BCL2 at score 3.16 (p=0.0007). With this change it returns 15 matches.

`--max-pvalue` still works as a deprecated alias. It filters on the whole-query p-value only, reproducing the old behavior, and prints a warning. If `--min-region-score` is also passed explicitly alongside `--max-pvalue`, a second warning explains that the region cap is being ignored rather than silently dropping it.

One cost: region-scope filtering can no longer be checked before `find_matched_regions` runs, because whether a pair passes now depends on the regions that function finds. Pairs that fail the whole-query check still pay the cost of region-finding before being rejected.

## Performance

`region_expectation` (the function that computes `region_expected_shared_kmers`/`lambda_region` above) originally rescanned every one of the query's k-mers for every matched region. Since `compare()` runs once per candidate target for a given query, and rescopes the Poisson test to every one of that pair's matched regions, a query with several regions across many targets was rescanning its full k-mer set target-count x region-count times, even though a region only spans a handful of residues. Fixed by precomputing a per-query prefix-sum array once (`PreparedQuery::position_prefix`, built where the query is already prepared once before the per-target loop), so `region_expectation` is now an O(1) prefix-sum lookup per region instead of an O(query k-mer count) rescan. Not benchmarked at wall-clock level -- this repo's test databases (25 targets) are too small for the difference to clear noise -- but the complexity improvement is real and compounds with database size.

## Not included

A region's containment ratio (shared k-mers over the region's own k-mers) is always 1.0, because a region is defined as a gapless run, so every k-mer in its span is already shared by construction. A separate count of "how many query k-mers fall in the region" would be identical to the shared count for the same reason, so neither is reported.

`db_n_residues` (total amino acids in the database) is not available: `ProteomeIndexMetadata` does not store sequence lengths, and adding it needs an index format change plus a reindex. `db_n_kmers` (total k-mer occurrences) is reported instead, as the closest available substitute.

## visualize_hits.py

This script crashed with `KeyError: 'poisson_pvalue'` on real output from this branch. Its test fixture still used the old column names, so all of its tests passed while the script itself was broken against every real CSV. The fixture is now generated to match `SearchResultCsv` directly, checked by a schema test that parses the Rust struct, so this kind of drift shows up as a test failure instead of passing silently.

Its Benjamini-Hochberg FDR correction, which produces the q-value used to rank hits, now runs over each target's best-scoring region instead of `query_poisson_pvalue`. `benjamini_hochberg()` expects p-values (smaller is more significant); the script reads `region_tail_probability` directly off the same CSV row it takes the score from, rather than reconstructing a probability from the score (`10 ** -score`). Correcting the whole-query number instead would push every region-driven hit to a q-value near 1, and `--max-hits` would then cut the hits this feature exists to surface. Display text changed from "region p=" to "region score=" to match.

## Verification

- 148 Rust tests, 54 Python tests, all passing.
- Ran the CLI end-to-end on real data. The E-value shown above, and every number in this description, was measured from the written CSV and rendered plot, not asserted only in a unit test.
- Confirmed parquet conversion carries all new columns and no longer includes the old `region_bonferroni_n` column.
- Inspected the rendered plot: the BCL2 landmark region appears and ranks correctly (`region score=3.2`, `region q=7.2e-08` for the strongest hit in this database, `query p=0.99` for the diluted whole-protein match).
- Checked `poisson_survival`'s tail direction against statrs's pinned source (`statrs-0.17.1`): `Poisson::cdf(x)` computes `P(X <= x)`, so `1.0 - cdf(n - 1)` is `P(X >= n)`, the intended "at least this many" test. Not an off-by-one.
- Added `debug_assert!` at both places `region.length - ksize + 1` is computed, so a region shorter than `ksize` (which should be structurally impossible; see `find_matched_regions`) fails loudly in debug and test builds instead of `saturating_sub` silently returning `n_shared = 1`.

Out of scope: notebooks 084/123/125/126/200–206 in `2024-kmerseek-analysis` still reference the old column names.


scripts/test_visualize_hits.py
```diff
@@ -104,34 +106,83 @@ def test_resolve_query_names_no_match_returns_empty():
...
-        "jaccard": str(jaccard),
-        "enrichment": str(enrichment),
-        "poisson_pvalue": str(poisson_pvalue),
+        "n_intersecting_hashes": "5",
+        "ksize": "10",
+        "scaled": "1",
         "moltype": moltype,
-        "query_subseq": query_subseq,
-        "moltype_seq": moltype_seq,
+        "jaccard": str(jaccard),
```
olgabot: Why are all these number fields strings? Aren't they integers and floats?

src/rust/search.rs
```diff
@@ -31,18 +31,40 @@ pub struct SearchFilters {
...
+    }
+}
+
+impl SearchFilters {
+    /// A pair is kept when *either* scope clears its cap, not both.
+    ///
+    /// AND would reintroduce whole-protein dilution through the back door: BCL2/CED9 at k=15
+    /// has a whole-query p of 0.99 and a region p of 0.0007, so requiring both to pass discards
+    /// exactly the sub-protein domain calls region scoring exists to surface.
+    fn pvalues_pass(&self, query_pvalue: f64, best_region_pvalue: Option<f64>) -> bool {
```
olgabot: I'd want the region p-value to be preferred... I want to see examples where the whole protein passes but the regio9n doesn't

src/rust/search.rs
```diff
@@ -257,13 +318,84 @@ pub struct MatchedRegion {
...
+    } else {
+        0.0
+    }
+}
+
+/// Fraction of `encoded` that is not its most common character.
+///
+/// Generalizes minority fraction to any alphabet: for 2-letter HP it is exactly
+/// `min(count_h, count_p) / len`, and for dayhoff it reads as the non-modal fraction.
+fn non_modal_fraction(encoded: &str) -> Option<f64> {
```
olgabot: Don't do this, removing low complexity in a separate PR: https://github.com/seanome/kmerseek/pull/37#pullrequestreview-4897405114

src/rust/search.rs
```diff
@@ -957,6 +1152,14 @@ fn calculate_similarity_from_precomputed(
 
     let matched_regions = find_matched_regions(query, target, intersection);
 
+    // Number of distinct positions a region could have started at in this query. Regions are
+    // selected by maximization over these placements, so this is the multiplicity that a
+    // per-region p-value has to be corrected against.
+    let region_search_space = query
+        .get_raw_sequence()
+        .map(|seq| seq.len().saturating_sub(query.protein_ksize() as usize) + 1)
```
olgabot, line 1293: what does saturating_sub help with here?

src/rust/search.rs
olgabot: Use human-mouse gencode genes as an example

src/rust/search.rs
```diff
@@ -1339,37 +1573,37 @@ mod tests {
...
-            query_start: 267,
-            query_end: 280,
+            start: 267,
+            end: 280,
             target_start: 200,
             target_end: 213,
         },
         landmark: Some(ExpectedMatchedRegion {
-            query_subseq: "QCPMSYGRLIGLISFGGFV",
+            subseq: "QCPMSYGRLIGLISFGGFV",
```
olgabot, line 1736: Yay thank you fo9r including this!

src/rust/search.rs
```diff
@@ -2361,4 +2655,374 @@ mod tests {
...
+        Ok(())
+    }
+
+    /// The multiplicity components are reported separately rather than pre-multiplied into the
+    /// p-value, so a hit's significance never shifts with batch composition. Checks each
+    /// component counts what it claims, and — the property that motivates the split — that
+    /// searching the same query alone leaves its p-values and region_search_space untouched
+    /// while only run_n_queries moves.
+    #[test]
+    fn test_multiplicity_components_are_reported_separately() -> Result<()> {
```
olgabot, line 2941: I don' tunderstand what this is testing

src/rust/search.rs
```diff
@@ -2361,4 +2655,374 @@ mod tests {
...
+            assert!(pair[0].containment >= pair[1].containment, "results must stay sorted");
+        }
+
+        Ok(())
+    }
+
+    /// minority_fraction flags regions whose encoded sequence is compositionally skewed, where
+    /// region enrichment looks most impressive and means least. It is the non-modal fraction, so
+    /// a homopolymer run is 0.0 and an evenly mixed HP region approaches 0.5.
+    #[test]
```
olgabot, line 3175: what d9oes non-modal mean?

src/rust/search.rs
olgabot: Explain multiplicity clearly and succinctly

src/rust/search.rs
```diff
@@ -743,11 +902,39 @@ impl ProteinSearcher {
...
-            0.0
-        };
+        let query_enrichment =
+            fold_enrichment(n_intersecting_hashes as u32, query_expected_shared_kmers);
+
+        // Rescope the same Poisson test to each matched region individually, so a tight local
+        // match doesn't get diluted by the whole protein's k-mer count.
+        let ksize = query.sketch.protein_ksize() as usize;
+        for region in result.matched_regions.iter_mut() {
+            let lambda = self.region_expectation(query.sketch, region.start, region.end, ksize);
```
olgabot: Specify that lambda is the sum of database abundances of the k-mers from the region, divided by the region size, in the code and the PR text. It is not clear now

src/rust/search.rs
```diff
@@ -743,11 +902,39 @@ impl ProteinSearcher {
...
+            // broken, instead of saturating_sub silently turning a bug into n_shared = 1.
+            debug_assert!(
+                region.length >= ksize as u32,
+                "region shorter than ksize: length={}, ksize={ksize}",
+                region.length
+            );
+            let n_shared = region.length.saturating_sub(ksize as u32) + 1;
+
+            region.expected_shared_kmers = lambda;
+            region.poisson_pvalue = poisson_survival(n_shared, lambda);
```
olgabot: The PR text says this isn't a real pvalue due to lack of independence. So can we call it something else?

scripts/visualize_hits.py
```diff
@@ -547,7 +586,11 @@ def _render_query(query_name, query_rows, query_length, args):
     target actually tested, not just the ones that end up displayed, or it
     would understate how many comparisons were made.
     """
-    corrected_pvalues = benjamini_hochberg(_target_pvalues(query_rows))
+    # benjamini_hochberg expects p-values (smaller is more significant), but
+    # _target_scores returns -log10 scores (bigger is more significant), so
+    # convert back before correcting.
+    target_pvalues = {name: 10.0**-score for name, score in _target_scores(query_rows).items()}
```
olgabot: Can we just keep both the score and original pvalue so we don't have to do this conversion? Maybe call it not a pvalue but something else?

src/rust/search.rs
```diff
@@ -771,14 +984,56 @@ impl ProteinSearcher {
...
+    fn region_expectation(&self, query: &ProteinSketch, start: u32, end: u32, ksize: usize) -> f64 {
+        let window_start = start as usize;
+        // A k-mer at p covers [p, p + ksize), so it fits inside [start, end) only when
+        // p <= end - ksize. A span shorter than ksize holds no whole k-mer at all - saturating
+        // here would wrongly admit position 0.
+        let window_end = match (end as usize).checked_sub(ksize) {
+            Some(last_start) => last_start + 1,
+            None => window_start,
+        };
+        query
```
olgabot: Is this the most efficient iteration? What kind of time are we losing here?

olgabot (commented):
This is almost ready

scripts/visualize_hits.py
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Done: `MatchedRegion` and the CSV now carry both `poisson_score` (the -log10 score, for display/ranking) and a new `tail_probability` field (the raw Poisson survival probability, for anything that needs a real probability). Named `tail_probability`, not `poisson_pvalue` — it has the same two structural problems documented on `poisson_score`, so calling it a p-value would be the same overclaim we already renamed away from.

`visualize_hits.py`'s Benjamini-Hochberg correction now reads `region_tail_probability` directly off the same row it takes the score from, instead of computing `10**-score`. See `be041cf`.

src/rust/search.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Good catch, this was genuinely wasteful. `region_expectation` was rescanning every one of the query's k-mers (`query.kmer_positions()`, a `HashMap` over the whole protein) for every matched region, and `compare()` is called once per candidate target per query, so a query with several regions across many targets was rescanning its full k-mer set target-count x region-count times.

Fixed by precomputing a per-query prefix-sum array once (`PreparedQuery::position_prefix`, built in `prepare_query`, which already runs once per query before the per-target loop), so `region_expectation` is now a prefix-sum difference: O(1) per region lookup instead of O(query k-mer count) per region. See `be041cf`.

I didn't benchmark wall-clock time — the repo's test databases are only 25 targets, too small for the difference to be measurable over noise. The complexity improvement is real and compounds with database size; happy to run a proper benchmark against a larger database if that's useful before merging.

PR #36, opened 2026-08-06
Log k-mer frequency histogram and top/bottom 10 k-mers during indexing

olgabot:
## Summary
During indexing, log the k-mer frequency distribution to stderr so the shape of an index is visible without a separate analysis pass, and optionally save it as CSV for plotting.

- A power-of-two-binned histogram of k-mer occurrence counts, with the total and unique k-mer counts, the mean, and the median.
- The 10 most and 10 least common k-mers, shown as encoded k-mer strings rather than hashes. Hashes are one-way murmur values, so each is resolved back to its text by looking up a signature containing it and slicing the stored sequence at the recorded position.
- `--kmer-stats-out PATH` writes the frequency spectrum as CSV, gzipped when the path ends in `.gz`.

Each k-mer is counted once per sequence containing it, so "total" is the number of (sequence, k-mer) pairs rather than a count of k-mer positions.

## Example output

```
[save] K-mer frequency histogram: 164269039 total k-mers found, 16711823 unique (mean 9.83, median 8.0 sequences per k-mer)
[save]              1 occurrences:     490084 k-mers  #
[save]            2-3 occurrences:    1119750 k-mers  ##
[save]            4-7 occurrences:    4181835 k-mers  ##########
[save]           8-15 occurrences:   15943791 k-mers  ########################################
...
[save] 10 most common k-mers (encoded k-mer: occurrences):
[save]   hhhhhhhhhhphphhhhphhpphh: 1534
[save]   hhhhhhhphphhhhphhpphhphh: 1501
...
[save] 10 least common k-mers (encoded k-mer: occurrences):
[save]   hphpphXhph: 1
[save]   (2485879 k-mers occur 1x; showing an arbitrary sample)
```

The CSV carries the totals in a leading `#` comment rather than a column, since they are constant for the file and would otherwise repeat on every row. Readers skip it with a comment prefix, for example `pl.read_csv(f, comment_prefix="#")`.

```
# total_kmers=164269039 unique_kmers=16711823 mean_seqs_per_kmer=9.8295 median_seqs_per_kmer=8.0 moltype=hp ksize=24
moltype,ksize,occurrences,n_kmers
hp,24,1,490084
hp,24,2,712205
```

`moltype` and `ksize` repeat per row so files from different runs concatenate into one frame.

## SwissProt sweep

[...]

PR #40, opened 2026-08-17
Build the k-mer spectrum once and propagate gzip finish errors

olgabot:
Follow-up to #36, which merged before these review fixes were applied. No behaviour changes.

## Truncated gzip files were reported as success

`write_kmer_frequency_spectrum` boxed the encoder as `Box<dyn Write>`, so `finish()` was never called and the gzip trailer was left to `Drop`. flate2 1.1.2 discards the error there:

```rust
// flate2-1.1.2/src/gz/write.rs:161
impl<W: Write> Drop for GzEncoder<W> {
    fn drop(&mut self) {
        if self.inner.is_present() { let _ = self.try_finish(); }
    }
}
```

A failed final write left a truncated `.gz` on disk while the function returned `Ok(())`. This is reachable: the disk filled during the SwissProt sweep that produced the current spectra. The encoder is now finished explicitly so the error propagates, which meant splitting the CSV body into a generic `write_spectrum_csv<W: Write>` instead of writing through a trait object.

All 134 existing spectra pass `gzip -t`, so nothing already written is damaged.

## Nine passes over the frequency map

The summaries walked `kmer_frequencies` nine times: `frequency_bins`, `values().sum()` and `frequency_spectrum` for the histogram; `frequency_spectrum` and `values().sum()` again for the CSV; two `n_smallest_by_key` calls; and a `values().filter(..).count()` tie count per example list. That map holds 125,883,747 entries at protein k=15.

The spectrum is built once and everything else derives from it:

- `total_kmers` is `sum(occurrences * n_kmers)`
- `unique_kmers` is `sum(n_kmers)`
- `frequency_bins` buckets the spectrum by `leading_zeros` instead of rescanning
- the tie count is a `BTreeMap` lookup rather than a full scan

Only the two heap selections still touch the map, since the spectrum discards hashes. Nine passes down to three.

### Measured

Two release binaries differing only in the summary code, each timing the block around `log_kmer_frequency_stats`, alternating old and new per repetition on the same input so machine drift hits both equally.

| input | unique k-mers | old | new | saved | speedup |
|---|---|---|---|---|---|
| SwissProt, protein k=15 | 125,883,747 | 5,859 ms | 1,404 ms | 4,455 ms | 4.2x |
| 250k subset, protein k=9 | 53,687,921 | 2,310 ms | 522 ms | 1,788 ms | 4.4x |
| SwissProt, hp k=24 | 16,711,823 | 1,204 ms | 354 ms | 850 ms | 3.4x |

Run-to-run spread was under 1% (protein k=15 measured 5,849 and 5,870 ms old, 1,406 and 1,403 ms new).

For context, a full protein k=15 index of SwissProt takes 235 s, so the summary block goes from 2.5% to 0.6% of the run, saving 4.5 s per index. Across the 134-run sweep that is roughly 10 minutes. Indexing is still dominated by hashing and this does not change that.

## Smaller items

- `n_smallest_by_key` replaces the heap's worst entry through `peek_mut` instead of pushing and immediately popping for every element that cannot make the list, which is nearly all of them.
- `frequency_spectrum` carries a `debug_assert!` that counts are non-zero. `frequency_bins` computes `usize::BITS - count.leading_zeros() - 1`, which underflows at zero. Callers build counts by incrementing from zero so this holds, but it was an unstated invariant that only private helpers and tests could break.
- `median_occurrences` derives `unique` from the spectrum instead of taking it as a second argument that could disagree with it.
- `print_frequency_histogram` uses a `let ... else` on the bin range rather than `unwrap()`.

## Test plan
- [x] `cargo test --no-default-features --lib -- --test-threads=1` — 136/136 pass
- [x] `cargo fmt --check` and `cargo clippy --lib` clean
- [x] Re-indexed all 572,970 SwissProt sequences at hp k=24 and diffed the spectrum CSV against the one built before the refactor: identical
- [x] `gzip -t` passes on all 134 spectra from the sweep

The protein spectra are not byte-identical between the two versions, but neither are two runs of the *same* binary: `main` resolves B, J and Z by drawing at random, so which k-mers appear shifts every run. Indexing a 60,000-sequence slice at protein k=9 four times gave 13,800,139, 13,800,133, 13,800,123 and 13,800,137 unique k-mers. `total_kmers` was identical at 21,208,286 in all four, so the count of (sequence, k-mer) pairs is stable and only their identity moves. #39 fixes this. It affects both binaries equally, so the timings above are unaffected.

Note for reviewers: `cargo clippy --all-targets` fails on `main` because `benches/benchmark.rs:455,463,529` call `searcher.search_one(&query_sig)` while `search_one` has taken a second `&SearchFilters` argument since #27. Unrelated to this PR, and CI reports the Benchmark job as `skipping`, so it stays hidden.

### DIFF
--- a/src/rust/index.rs
+++ b/src/rust/index.rs
@@ -3578,6 +3578,12 @@ mod tests {
         assert_eq!(least, vec![(200, 1), (100, 2)]);
     }
 
+    /// Real N-terminal fragment of C. elegans CED-9 (UniProt P41958) with three residues
+    /// rewritten to the ambiguity codes that stand for them: Asn->B (Asx), Glu->Z (Glx),
+    /// Ile->J (Xle). Also carries U (Sec) and O (Pyl).
+    const CED9_WITH_AMBIGUITY_CODES: &str =
+        "MTRCTADNSLTNPAYRRRTMBTGEMKEFLGJKGTEPTDFGZNSDAQDLPSPSRQASTRRUO";
+
     /// Indexing the same sequence twice must produce byte-identical sketches.
     ///
     /// WHY: ambiguity codes were previously resolved by drawing at random from the
--- a/src/rust/index.rs
+++ b/src/rust/index.rs
@@ -379,9 +379,13 @@ impl ProteomeIndex {
         if kmer_frequencies.is_empty() {
             return Ok(());
         }
-        Self::print_frequency_histogram(kmer_frequencies);
+        // Build the spectrum once. Every summary below (bins, total, unique, median, tie
+        // counts) is derived from it rather than rescanning the map, which holds over
+        // 125 million entries for protein k=15.
+        let spectrum = Self::frequency_spectrum(kmer_frequencies);
+        Self::print_frequency_histogram(&spectrum);
         if let Some(path) = kmer_stats_out {
-            self.write_kmer_frequency_spectrum(path, kmer_frequencies)?;
+            self.write_kmer_frequency_spectrum(path, &spectrum)?;
         }
 
         // Resolving a hash back to text needs the sequence it came from, which is only in
@@ -397,25 +401,12 @@ impl ProteomeIndex {
         });
         let least = Self::n_smallest_by_key(kmer_frequencies, n, |hash, count| (count, hash));
 
-        self.print_kmer_examples(
-            "most common",
-            &most,
-            kmer_frequencies,
-            inverted_index,
-            target_list,
-        );
-        self.print_kmer_examples(
-            "least common",
-            &least,
-            kmer_frequencies,
-            inverted_index,
-            target_list,
-        );
+        self.print_kmer_examples("most common", &most, &spectrum, inverted_index, target_list);
+        self.print_kmer_examples("least common", &least, &spectrum, inverted_index, target_list);
         Ok(())
     }
 
-    /// Write the k-mer frequency spectrum (occurrences -> how many k-mers had that many)
-    /// as CSV, gzip-compressed when the path ends in `.gz`.
+    /// Write the k-mer frequency spectrum as CSV, gzip-compressed when the path ends in `.gz`.
     ///
     /// WHY the spectrum rather than one row per k-mer: this is the distribution you plot, and
     /// it is a few thousand rows instead of tens of millions, so a sweep over alphabets and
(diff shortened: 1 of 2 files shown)


src/rust/index.rs
```diff
@@ -334,12 +334,12 @@ impl ProteomeIndex {
             inverted_index.len(),
         );
 
-        self.log_kmer_frequency_stats(
-            &kmer_frequencies,
-            &inverted_index,
-            &target_list,
-            kmer_stats_out,
-        )?;
+        self.log_kmer_frequency_stats(&kmer_frequencies, kmer_stats_out, |hashes| {
```
olgabot, line 474: Add a comment explaining the reason for this

src/rust/index.rs
```diff
@@ -397,25 +404,17 @@ impl ProteomeIndex {
...
-            target_list,
-        );
-        self.print_kmer_examples(
-            "least common",
-            &least,
-            kmer_frequencies,
-            inverted_index,
-            target_list,
-        );
+        let mut wanted: Vec<u64> = most.iter().chain(least.iter()).map(|&(hash, _)| hash).collect();
```
olgabot: I don't like this variable name, "wanted" Unclear WHY this is "wanted"

src/rust/index.rs
```diff
@@ -522,22 +548,22 @@ impl ProteomeIndex {
...
+        };
+        let max_count = bins.values().max().copied().unwrap_or(1);
+
+        let total = Self::total_kmers(spectrum);
+        let unique = Self::unique_kmers(spectrum);
         eprintln!(
             "[save] K-mer frequency histogram: {total} total k-mers found, {unique} unique \
              (mean {:.2}, median {:.1} sequences per k-mer)",
             total as f64 / unique as f64,
-            median,
```
olgabot, line 674: let's add mode as well

olgabot (commented):
Needs a few more changes, like comments, variable renames, and add the mode as well as mean and median

src/rust/index.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Added a comment at the call site explaining why the closure is used instead of the raw `inverted_index`/`target_list`: `save_kmer_stats_only` needs to plug in a different, lower-memory resolution strategy. See 0244494.

src/rust/index.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Renamed `wanted` to `example_hashes` and added a comment explaining the dedup (a hash landing in both `most` and `least` when there are very few unique k-mers). See 0244494.

src/rust/index.rs
olgabot: > 🤖 **This reply was written by [Claude Code](https://claude.com/claude-code)**, not by @olgabot — it is posted from her account because that is the authenticated `gh` login.

Added mode alongside mean and median, both in the console histogram summary and the CSV totals comment, plus a test covering the tie-break rule. See 0244494.

PR #42, opened 2026-08-19
Add hp_lehninger_hpc: 3-letter hydrophobic/polar/cystine alphabet

olgabot:
For capturing C2H2 Zinc finger proteins, the cystines have a special role in holding the structure together, so we want to add this alphabet to see if we can detect them well.

Extends hp_lehninger_plus_c by giving cysteine its own symbol ('c') instead of folding it into hydrophobic, since disulfide-bond formation is a distinct chemistry from ordinary hydrophobic packing. Wires the new hp_lehninger_hpc moltype through the CLI, encoding validation, and tests.

### DIFF
--- a/src/rust/tests/test_hp_encoding.rs
+++ b/src/rust/tests/test_hp_encoding.rs
@@ -139,7 +139,7 @@ mod tests {
             "hp_thomas_dill",
             "hp_lehninger",
             "hp_thomas_dill_no_c",
-            "hp_lehninger_plus_c",
+            "hp_lehninger_c_nonpolar",
             "hp_lehninger_hpc",
             "hp_pbotc_1st_ed",
         ];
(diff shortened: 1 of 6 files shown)


src/rust/tests/test_hp_encoding.rs
```diff
@@ -140,6 +140,7 @@ mod tests {
             "hp_lehninger",
             "hp_thomas_dill_no_c",
             "hp_lehninger_plus_c",
```
olgabot: Let's rename "hp_lehninger_plus_c" because it's confusing compared to hp_lehninger_hpc. How about "hp_lehninger_c_nonpolar"?

olgabot (commented):
The naming is confusing

src/rust/tests/test_hp_encoding.rs
olgabot: Renamed to `hp_lehninger_c_nonpolar` in d60a87e. — Claude

PR #43, opened 2026-08-22
Add reduced amino acid alphabets from Peterson 2009 and Ieremie 2024, and name every alphabet by its class count

olgabot:
Adds the reduced amino acid alphabets from four papers, renames every alphabet so its
name states how many classes it collapses the 20 residues into, and changes how
ambiguity codes are indexed.

Sources:
- Peterson et al. (2009), [doi:10.1093/bioinformatics/btp164](https://doi.org/10.1093/bioinformatics/btp164), Table 2
- Ieremie et al. (2024), [doi:10.1093/bioinformatics/btae061](https://doi.org/10.1093/bioinformatics/btae061), Table 1
- Rannon & Burstein (2026)
- Jain, Jain & Jain (2014); Ball, Hill & Scott (2014)

## Alphabets

Nineteen partitions in `src/rust/alphabets.rs`, 2 to 20 classes, as one `Alphabet` enum:

| Alphabet | Classes | Source |
|---|:---:|---|
| `hp_lehninger2` … `hp_pbotc_1st_ed2` | 2 | six HP schemes, differing only on C, G, P, W and Y |
| `hp_lehninger_hpc3` | 3 | Lehninger with C in a class of its own |
| `gbmr4` | 4 | Solis & Rackovsky 2000. Best recall at 0.01 EPQ in Peterson et al. |
| `polarity4` | 4 | Ball, Hill & Scott 2014. Lowest stability-regression RMSE in Rannon & Burstein |
| `wwmj5` | 5 | Wang & Wang 1999, Miyazawa-Jernigan contact potentials |
| `dayhoff6` | 6 | Dayhoff |
| `gbmr7` | 7 | Solis & Rackovsky 2000 |
| `funcgroups8` | 8 | Jain, Jain & Jain 2014. Best solubility AUROC in Rannon & Burstein |
| `sdm12` | 12 | Prlic et al. 2000. Best AUC in Peterson et al. |
| `mmseqs12` | 12 | Steinegger & Soding 2018 |
| `wass14` | 14 | Ieremie et al. 2024, hydrophobicity-clustered |
| `hsdm17` | 17 | Prlic et al. 2000. Best mean pooled precision in Peterson et al. |
| `uniprot18` | 18 | Ieremie et al. 2024, clusters learned by a protein language model |
| `protein20` | 20 | no reduction |

GBMR4, SDM12 and HSDM17 appear in both Peterson et al. and Ieremie et al. with identical
clusters, which cross-checks the transcription. `mmseqs12` appears in Rannon & Burstein
under its Linclust name.

Each class is written as its first residue in lowercase, so an SDM12-encoded sequence
shows the `LIVM` class as `l` and can be read against the source residues.
`build_reduced()` asserts full coverage of the 20 canonical residues and that no two
classes share a leading residue.

## Naming

The flag is `--alphabet` (`-a`), and every value ends in its class count:

| Alphabet | Was |
|---|---|
| `protein20` | `protein`, `raw` |
| `dayhoff6` | `dayhoff` |
| `hp_lehninger2` | `hp_lehninger`, `hp` |
| `hp_thomas_dill2` … `hp_lehninger_hpc3` | `hp_<name>` |
| `gbmr4` … `uniprot18` | new |

An earlier round of this branch prefixed everything with `reduced_`. That carried no
information, since every alphabet except the full one is reduced, and
`reduced_hp_thomas_dill2` stated its class count twice over (`hp` and `2`). Dropping it
leaves `<name><classes>`, the convention the source papers use: Peterson et al. Table 1
lists schemes as TD, WW, GBMR, HSDM, SDM.

The `hp_` prefix marks the alphabets that differ only on borderline residues, which keeps
a sweep easy to write and to grep. `polarity4` and `funcgroups8` have no name in their
source, so they are named for what they group on.

This is a clean break: only the current names parse. `--encoding` and its value aliases
are gone.

## Amino acid disambiguation

Three one-letter codes stand for a pair of residues rather than a single one, because the
method that produced the sequence could not tell the pair apart. Asn and Gln deamidate to
Asp and Glu during acid hydrolysis, and Ile and Leu have the same mass:

| code | name | stands for |
|---|---|---|
| `B` | Asx | `D` (Asp, aspartate) or `N` (Asn, asparagine) |
| `J` | Xle | `I` (Ile, isoleucine) or `L` (Leu, leucine) |
| `Z` | Glx | `E` (Glu, glutamate) or `Q` (Gln, glutamine) |

kmerseek indexes every k-mer covering one of these under both readings, so a query
holding either residue matches. It does not pick one: under `sdm12` and `hsdm17`, Asp and
Asn fall in different classes, so choosing `D` for a `B` would assert a residue the source
never had.

Whether disambiguating adds k-mers depends on the alphabet, because the two readings do
not always encode differently. Under `dayhoff6` Asp and Asn are both class `c`, and under
the HP tables both are polar, so the two readings of a window produce the same encoded
k-mer and the same hash. Under `protein20`, `sdm12` and `hsdm17` they encode differently,
so the window yields two k-mers instead of one.

`PLANTANDANIMALGENBMES` is 21 residues with a `B` at index 17, so at k=5 it has 17
windows and four of them span the `B`:

| alphabet | k-mers with `B` resolved to one residue | k-mers with `B` disambiguated |
|---|---|---|
| `protein20` | 17 | 21 |
| `sdm12` | 17 | 21 |
| `hsdm17` | 17 | 21 |
| `dayhoff6` | 17 | 17 |
| `hp_lehninger2` | 14 | 14 |

`protein20`, `sdm12` and `hsdm17` keep `D` and `N` in different classes, so each of the
four `B`-windows becomes two k-mers: 17 + 4 = 21. `dayhoff6` puts both in class `c`, so
the four windows stay one k-mer each. `hp_lehninger2` starts at 14 rather than 17, for a
reason unrelated to the `B`: with only two symbols, three pairs of ordinary windows
already encode identically.

A window may be at most 10% ambiguity codes, meaning `ceil(ksize / 10)` of them. A window
with more is dropped: indexing only part of its readings would make matching depend on
which subset was kept, which is worse than losing that one window. SwissProt holds about
900 non-canonical residues in 207.6 M, so a window over the cap should not arise.

The cap was a flat four codes for most of this branch, which was tied to the reading count
it produced rather than to how ambiguous the window actually was: four codes is 80% of a
k=5 window and 8% of a k=50 one. Scaling with `ksize` ties the 2^*n* growth to the fraction
of the window that is ambiguous. Rounding up keeps a single code legal at every k-size.

`U` (Sec, selenocysteine) and `O` (Pyl, pyrrolysine) are handled differently. They are
specific residues rather than ambiguities, so each takes its closest canonical analogue,
`C` and `K`, under a reduced alphabet. Under `protein20` they are kept as themselves.

## sourmash compatibility

sourmash calls an alphabet a moltype, and writes three of them. kmerseek reads all
three, storing each under the kmerseek name for the same alphabet:

| sourmash | kmerseek | hash function |
|---|---|---|
| `protein` | `protein20` | `Murmur64Protein` |
| `dayhoff` | `dayhoff6` | `Murmur64Dayhoff` |
| `hp` | `hp_lehninger2` | `Murmur64Hp` |

These three are the alphabets sourmash encodes itself, so kmerseek hands the sequence
straight to it rather than pre-encoding. A sketch or index carrying a sourmash name
therefore holds hashes kmerseek can read as-is, on the command line and from stored
index metadata:

```
Alphabet: HpLehninger (detected: hp)
Total matches: 21
```

## Indexes built before this change

kmerseek's own earlier spellings are not read, since they were never sourmash names and
so have nothing to stay compatible with: `raw`, and `hp_<name>` without its class count.
An index recording one fails to open:

```
Unknown alphabet in database: hp_thomas_dill
```

No hash function changed, so rebuilding under the current name gives the same hits.

## Choosing a class count

Class count and k-size trade off against each other, so a k that suits a 2-class alphabet
is usually too long for a finer one. Searching CED-9 against the 25-sequence BCL-2 test
file:

| Alphabet | k=10 targets / matched regions | k=5 targets / matched regions |
|---|---|---|
| `hp_lehninger2` | 21 / 1673 | none |
| `gbmr4` | 21 / 294 | 14 / 22234 |
| `polarity4` | 14 / 153 | 15 / 9884 |
| `funcgroups8` | 2 / 3 | 18 / 982 |
| `sdm12` | none | 17 / 110 |

At k=10, `gbmr4` finds the same 21 targets as `hp_lehninger2` in a fraction of the matched
regions, the selectivity gain Peterson et al. describe, while the finer alphabets have
almost nothing left. At k=5 that reverses.

## Verification

- 225 Rust tests, 54 Python tests, `cargo fmt --check` clean, no new clippy warnings.
- Every cluster constant and its expected encoding was checked programmatically against
  the papers rather than by eye. That caught two transcription errors during development:
  GBMR7's F, and a transposition in WWMJ5.
- The CLI index test covers all 19 alphabets, built from `Alphabet::all()` so a new
  alphabet is covered without editing the test, with `assert_eq!(alphabets.len(), 19)` so
  the list cannot silently shrink.
- A test pins the README's HP table: `AFILMV` hydrophobic and `DEHKNQRST` polar in every
  scheme, with only C, G, P, W and Y free to move.
- `test_hsdm17_refines_sdm12_refines_gbmr4` pins that going from GBMR4 to SDM12 to HSDM17
  only splits classes, never moves a residue across an existing boundary.
- Each of the three most recent commits builds and passes the suite on its own.
- The class-count table below was re-run against this branch's binary; both rows
  reproduce exactly.

## Notes for review

- The alphabet-size test compares every ordered pair rather than a sliding chain.
- `add_protein` tries the ambiguity branch before the table branch. With the order
  reversed, table-backed alphabets never expanded: `sdm12` gave 28 k-mers where the union
  is 33. The union assertion is what caught it.
- `query_tfidf` is compared with `approx` rather than `==`. It sums over a HashMap keyed
  by k-mer hash, so its last bits depend on the hash values.
- `hp_alphabets.rs` and `reduced_alphabets.rs` are now one `alphabets.rs`, and
  `encoding.rs` is `hash_functions.rs`, which is what it deals with. `HpAlphabet` and
  `ReducedAlphabet` are folded into one `Alphabet` enum.
- The random control alphabets are gone, along with `--random-seed` and the `rand`
  dependency. They were a negative control for the alphabet sweep, and the sweep does not
  need one badly enough to carry the machinery. Dropping the seeded variants also let
  `to_moltype()` go back to `&'static str`, since every alphabet now has a fixed name.
- `CED9_PREFIX` in `index.rs` is dead and `test_hp_encoding.rs` has an empty line after a
  doc comment. Both predate the recent commits and are left alone here.


src/rust/tests/test_cli.rs
```diff
@@ -84,7 +84,8 @@ fn test_cli_index_gzipped() -> Result<(), Box<dyn std::error::Error>> {
 fn test_cli_index_different_encodings() -> Result<(), Box<dyn std::error::Error>> {
     let temp_dir = tempdir()?;
 
-    for encoding in ["protein", "dayhoff", "hp"] {
```
olgabot, line 87: Test ALL encodings here, not just a few.

src/rust/tests/test_hp_encoding.rs
```diff
@@ -253,4 +253,69 @@ mod tests {
...
+    fn test_older_and_current_moltype_names_sketch_identically() {
+        use crate::hp_alphabets::HpAlphabet;
+
+        let seq = "TIFEKKHAENFETFCEQLLAVPRISFSLYQDVVRTVGNAQTDQCPMSYGRLIGLISFGGFV";
+
+        for alphabet in HpAlphabet::all_named() {
+            let current = alphabet.to_moltype();
+            // The two spellings this alphabet had before: no class count, and with the
+            // redundant `reduced_` prefix.
+            let older = [format!("hp_{}", alphabet.name()), format!("reduced_{current}")];
```
olgabot: Remove all of these `reduced_` tests, that's from the same PR ... nothing is "old" here

src/rust/tests/test_reduced_encoding.rs
```diff
@@ -0,0 +1,221 @@
+/// End-to-end tests for the multi-letter reduced alphabets in `crate::reduced_alphabets`.
```
olgabot: So what's the difference between encoding and reduced alphabets? Can we merge them? same with hp_alphabets? I guess hp is kind of its own special class but this is still confusing

src/rust/tests/test_reduced_encoding.rs
```diff
@@ -0,0 +1,221 @@
...
+                    8,
+                    1,
+                    &moltype,
+                )
+                .unwrap();
+                (moltype, sketch.mins_as_set())
+            })
+            .collect();
+
+        for (i, (left_name, left)) in sketches.iter().enumerate() {
```
olgabot, line 199: Shoulnd't this test all combos and not just a sliding pair?

src/rust/aminoacid.rs
```diff
@@ -76,27 +102,32 @@ impl AminoAcidAmbiguity {
     /// Validates a protein sequence, substituting representatives for non-canonical codes
     /// when `moltype` reduces the alphabet. Stops processing at the first stop codon (*).
     ///
-    /// WHY only for reduced, biochemically-derived alphabets: under Dayhoff or a named HP
+    /// WHY only for alphabets that collapse each ambiguous pair: under Dayhoff or a named HP
```
olgabot:
```suggestion
    /// WHY only for biochemically-derived alphabets that collapse each ambiguous pair: under Dayhoff or a named hydophobic-polar (HP)
```

src/rust/aminoacid.rs
```diff
@@ -76,27 +102,32 @@ impl AminoAcidAmbiguity {
...
     /// table a code like B encodes identically whether it is read as Asp or Asn, so
-    /// substituting a representative is lossless. Three moltypes are excluded:
-    ///   - `protein`/`raw` keep the full 20-letter alphabet, so there is no such equivalence
+    /// substituting a representative is lossless. Three groups of moltypes are excluded:
+    ///   - `protein20` (spelled `protein` or `raw` in older indexes) keeps all 20 residues, so there is no such equivalence
```
olgabot: Don't reference older indices.. nobody uses this right now

src/rust/aminoacid.rs
```diff
@@ -354,4 +385,58 @@ mod tests {
...
+        assert_eq!(
+            aa.validate_and_resolve("ACDEFXBZJUO", "gbmr4").unwrap().as_ref(),
+            "ACDEFXDEICK"
+        );
+    }
+
+    /// SDM12 gives Asp and Asn separate classes, so nothing is substituted and B/J/Z/U/O
+    /// hash as themselves, the same way X already does.
+    #[test]
+    fn test_validate_and_resolve_keeps_codes_verbatim_for_split_reduced_alphabets() {
```
olgabot: No don't do this -- add each kmer with both Asp and Asn for B

src/rust/encoding.rs
```diff
@@ -10,27 +14,36 @@ const MURMUR64HP: &str = "Murmur64Hp";
...
-             'hp_lehninger', 'hp_thomas_dill', 'hp_kyte_doolittle', \
-             'hp_thomas_dill_no_c', 'hp_lehninger_c_nonpolar', 'hp_lehninger_hpc', 'hp_pbotc_1st_ed', \
-             'hp_shuffled_control', 'hp_shuffled_control_1'..'hp_shuffled_control_10'",
+            "Invalid alphabet: {}. Supported values: 'protein20', 'dayhoff6', \
+             'hp_lehninger2', 'hp_thomas_dill2', 'hp_kyte_doolittle2', \
+             'hp_thomas_dill_no_c2', 'hp_lehninger_c_nonpolar2', 'hp_lehninger_hpc3', \
+             'hp_pbotc_1st_ed2', 'hp_random_control2', \
+             'hp_random_control2_1'..'hp_random_control2_10', \
+             'gbmr4', 'wwmj5', 'gbmr7', 'sdm12', 'mmseqs12', 'wass14', 'hsdm17', \
+             'uniprot18' (older spellings -- 'protein', 'raw', 'hp', 'dayhoff', \
```
olgabot: Don't reference "older spellings" let's make this a clean break

src/rust/sketch.rs
```diff
@@ -366,15 +366,16 @@ impl ProteinSketch {
...
-        use crate::hp_alphabets::HpAlphabet;
+        use crate::encoding::{
+            custom_alphabet_table, encode_by_moltype, encode_with_fn, get_encoding_fn_from_moltype,
+        };
         use crate::kmer::is_homopolymer_kmer;
         use sourmash::_hash_murmur;
 
         let moltype_str = self.moltype.to_string();
-        let custom_hp = HpAlphabet::from_moltype(&moltype_str);
+        let custom_table = custom_alphabet_table(&moltype_str);
```
olgabot: I don't like the name `custom_table`. Come up with something cleare

src/rust/sketch.rs
```diff
@@ -745,7 +712,7 @@ mod tests {
         // TEST_PROTEIN = "PLANTANDANIMALGENQMES"; the window at position 10,
         // "IMALG", is all-hydrophobic ("hhhhh") under the HP (Lehninger)
         // alphabet — the low-complexity case this targets.
-        const IMALG_HASH: u64 = 8541583772724823208;
+        const IMALG_HASH: u64 = 1279034388713273924;
```
olgabot: Actualy for hp_leninger2, let's keep it the same hashfunction as sourmash, for consistency

README.md
```diff
@@ -182,6 +256,53 @@ packing.
 | **W** | h | h | p | h | h | h | h |
```
olgabot: Let's show the HP alphbets like this, with the H andP constants in code tics to make it clear taht they're constant between different iterations. I think this is much easier to read than your table


Alphabet / Scheme | Hydrophobic (h) | Polar (p)
-- | -- | --
Lehninger (current) | `AFILMV` GPWY | `DEHKNQRST` C
Thomas-Dill/PBotC 2nd | AFILMV CWY | DEHKNQRST GP
Kyte-Doolittle | AFILMV C | DEHKNQRST GPWY
TD−C | AFILMV WY | DEHKNQRST CGP
Leh+C | AFILMV CGPWY | DEHKNQRST
PBotC 1st | AFILMV CPWY | DEHKNQRST G

olgabot (commented):
Needs changes

src/rust/tests/test_cli.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done. It now builds the list from `HpAlphabet::all_named()` and `ReducedAlphabet::all()` rather than hardcoding, with `assert_eq!(alphabets.len(), 18)` so the list cannot silently shrink — and a newly added alphabet is covered without touching the test.

src/rust/tests/test_hp_encoding.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

You're right, and I should have caught this. The `reduced_` prefix was introduced and removed inside this same branch, so those tests were pinning the PR against itself. Removed.

I took "clean break" to the rest of it too, per your comment on encoding.rs: `protein`, `raw`, `dayhoff`, `hp` and the count-less `hp_<name>` forms no longer parse either. `MolType::new` is back to validating rather than normalizing, and all the alias machinery is gone.

One consequence worth knowing: an index whose stored moltype is `hp` now fails with `Unknown alphabet in database: hp`. The *hashes* in it are fine now that Lehninger is back on sourmash's hash function — it is only the name that is unrecognized. Say the word if you want a one-line alias to make those readable.

src/rust/tests/test_reduced_encoding.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Fair — the split is not carrying its weight. Where it stands now:

- `hp_alphabets.rs` and `reduced_alphabets.rs` are the same concept: a partition of the 20 residues into classes, exposed as a `HashMap<u8, u8>`. The only real difference is that the HP ones share a family prefix and differ on five borderline residues.
- `encoding.rs` is a different thing: it picks the *hash function* and decides whether sourmash encodes the sequence or we pre-encode it. That distinction does still exist — `protein20`, `dayhoff6` and `hp_lehninger2` are encoded by sourmash, everything else by a table of ours — and it is now the single question `alphabet_table()` answers.

So I would merge the two alphabet modules into one `alphabets.rs` with one `Alphabet` enum, and keep `encoding.rs` as the sourmash boundary. That is a mechanical but wide refactor and I did not want to bury it inside a commit that also changes indexing behaviour. Want it as a follow-up commit on this branch, or a separate PR?

The two test files should merge along with them.

src/rust/tests/test_reduced_encoding.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done. It now compares every ordered pair, asserting that a coarser alphabet never uses more distinct symbols than a finer one, plus the per-alphabet bound that no alphabet exceeds its own class count. The sliding version could have hidden a mis-wired table between two neighbours that happened to be consistent.

src/rust/aminoacid.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Applied, though the doc comment it was attached to no longer exists in that form — with codes expanded rather than resolved, the "only for alphabets that collapse each ambiguous pair" caveat is gone. The rewritten comment says what `validate_and_resolve` still does (U and O only) and points at `ambiguity_readings` for B/J/Z. Also spelled out hydrophobic-polar (HP) as you suggested.

src/rust/aminoacid.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done — that whole paragraph is gone, along with the rest of the back-compat language across `aminoacid.rs`, `encoding.rs`, `types.rs` and the README.

src/rust/aminoacid.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done, and this was the right call — it fixes a real correctness gap rather than just a policy one.

`ambiguity_readings()` expands a k-mer window into every reading, and `add_protein` indexes all of them. A sequence carrying B now sketches to exactly the union of its Asp and Asn readings, asserted across protein20, dayhoff6, the HP tables, gbmr4, sdm12, hsdm17 and uniprot18.

Two details worth flagging:

- The expansion branch has to be tried **before** the table branch. I had it after at first, and table-backed alphabets silently never expanded — sdm12 gave 28 k-mers where the union is 33. Caught by comparing against the union rather than against one reading.
- Windows are capped at `MAX_AMBIGUITY_READINGS = 16` (four codes), and a window past it is skipped rather than indexed under an arbitrary subset. SwissProt has roughly 900 non-canonical residues in 207.6 M, so this should effectively never fire.

Whether the count grows depends on the alphabet, which the tests now state per alphabet: protein20 goes 17 -> 21 for a sequence with one B, while dayhoff6 stays at 17 and HP stays at 14 because both readings collapse to the same class there.

src/rust/encoding.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done — no aliases, no "older spellings" note in any error message. Only the current names parse. See the reply on the test_hp_encoding.rs thread for what that means for an existing `hp` database.

src/rust/sketch.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Renamed to `residue_classes`, which is what it holds — residue to class symbol. The function behind it is now `alphabet_table()` rather than `custom_alphabet_table()`; "custom" was doing no work since there is nothing non-custom to contrast with.

src/rust/sketch.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done, and this removes the hash incompatibility — good call.

`hp_lehninger2` is back on `Murmur64Hp`. Lehninger *is* sourmash's `aa_to_hp` partition, so its sketches are byte-identical to sourmash's. `alphabet_table()` returns `None` for it so sourmash does the encoding; every other HP table is still pre-encoded here, since none of them has a sourmash equivalent.

`ProteomeIndex::reject_legacy_builtin_hp` existed only to catch the mismatch this created, so it is gone, and the hp hash constants in the tests revert to their original values. Verified end to end: CED-9 against the 25-sequence BCL-2 file gives 21 matches and 1673 region rows, the same as before this branch.

**Correction to what I first wrote here:** I said an index built the old way needs no rebuild. That is wrong once the clean break is applied. The k-mers in it are still valid, since the hash function is the same, but its recorded moltype is `hp`, which no longer parses — so it fails with `Unknown alphabet in database: hp`. Only the name is the problem, and rebuilding gives identical hits.

README.md
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Agreed, that reads much better — the constant core is the point, and the old table buried it in 20 rows. Applied your layout, with `Leh HPC` split into its own small table since it has a third class.

Also added a test pinning the claim: `AFILMV` hydrophobic and `DEHKNQRST` polar in every scheme, with only C, G, P, W and Y free to move. Otherwise a new alphabet could quietly make the table wrong.

src/rust/tests/test_reduced_encoding.rs
olgabot: LEt's put all the alphabets into `alphabets.rs` and then rename `encoding.rs` -> `hash_functions.rs` since that's what it deals with

src/rust/sketch.rs
olgabot: So why is it `custom_alphabet_table`? Can't we name it `alphabet_table` then?

src/rust/sketch.rs
olgabot: Let's kep `hp` only parsing for compatibility with sourmash just in case

src/rust/tests/test_reduced_encoding.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done in `3a849f6`.

- `hp_alphabets.rs` + `reduced_alphabets.rs` → `alphabets.rs`. Both were the same thing, a partition of the 20 residues stored as a residue-to-class table, so the split was arbitrary.
- `encoding.rs` → `hash_functions.rs`.
- `alphabet_table()` moved from there into `alphabets.rs`, since it is a lookup over the alphabet catalog rather than anything to do with hash functions. That leaves a clean line: `alphabets.rs` owns the catalog, `hash_functions.rs` owns the sourmash boundary.

One thing I did not do: the two enums are still separate types inside `alphabets.rs`. Folding `HpAlphabet` and `ReducedAlphabet` into a single `Alphabet` enum is a further step, and it is not free — `HpAlphabet` carries the seeded `Random(n)` variants and `uses_sourmash_encoder()`, neither of which the multi-letter alphabets have. Say the word and I will do it as its own commit.

src/rust/sketch.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Already `alphabet_table` — I renamed it in the clean-break commit before you asked, so we agreed independently. "Custom" was contrasting with nothing.

The local variable in `add_protein` is `residue_classes`, which is what it holds.

src/rust/sketch.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done in `3a849f6`. Bare `hp` parses again, as `hp_lehninger2`.

This works cleanly now for the reason you gave earlier in this thread: since `hp_lehninger2` went back to sourmash's `Murmur64Hp`, the two names refer to the same partition **and** the same hash function, so a sourmash-labelled index holds hashes this alphabet can read. It is a real compatibility path, not just a name alias.

`MolType::new` normalizes `hp` to `hp_lehninger2` on the way in, so only one spelling reaches the rest of the code — `find_matched_regions` asserts query and target moltypes are equal, and two names for one alphabet would abort the search.

Verified on the index this branch was refusing an hour ago:

```
Alphabet: HpLehninger (detected: hp)
Total matches: 21
Total CSV rows: 1673
```

Same numbers as before the branch. Nothing else came back: `protein`, `dayhoff` and the count-less `hp_<name>` forms still fail, since those were kmerseek's own names and have no interop argument.

olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Extended to all three of sourmash's moltypes in `d9b3b60`, not just `hp` — same argument applies to `protein` and `dayhoff`, and I had scoped it too narrowly.

| sourmash | kmerseek | hash function |
|---|---|---|
| `protein` | `protein20` | `Murmur64Protein` |
| `dayhoff` | `dayhoff6` | `Murmur64Dayhoff` |
| `hp` | `hp_lehninger2` | `Murmur64Hp` |

These three are exactly the alphabets sourmash encodes itself, so kmerseek hands the sequence straight to it rather than pre-encoding. That is what makes the names safe to read: the hashes behind them are already kmerseek's hashes.

Replaced the `hp`-only special case with one `canonical_moltype()` in `alphabets.rs`, applied at each entry point (`MolType::new`, `alphabet_table`, both hash-function lookups), so there is a single place that knows about sourmash names.

One thing I got wrong on the first pass: the library accepted the names but clap did not, so `--alphabet protein` still failed while an index *recording* `protein` opened fine. Added the three as value aliases.

Verified end to end — all three parse on the CLI and store under kmerseek's name, an index written as `hp` gives its previous 21 matches, one written as `dayhoff` gives its previous 3.

Still rejected, since they were never sourmash names and have nothing to stay compatible with: `raw`, `hp_<name>` without a class count, and the `reduced_` prefix.

README.md
```diff
@@ -144,43 +141,206 @@ the others) -- use it to tame proteome-scale searches where a gene can have doze
...
+### Indexes built before this change
+
+kmerseek's own earlier spellings are not read, since they were never sourmash names and
+so have nothing to stay compatible with: `raw`, and `hp_<name>` without its class count.
+An index recording one fails to open:
+
+```
+Unknown alphabet in database: hp_thomas_dill
+```
+
```
olgabot:
```suggestion
```

README.md
```diff
@@ -144,43 +141,206 @@ the others) -- use it to tame proteome-scale searches where a gene can have doze
...
+  `hpphp`, `NTAND` and `NBMES` are both `pphpp`. Disambiguating adds none, so it stays
+  at 14.
+
+Disambiguating a code doubles the readings of every window it falls in, so a window
+holding *n* codes yields 2^*n* readings. kmerseek disambiguates a window holding at most
+4 codes, which is 16 readings. A window with more is dropped: indexing only part of its
+readings would make matching depend on which subset was kept, which is worse than losing
+that one window. SwissProt holds about 900 non-canonical residues in 207.6 M, so a window
+with five codes should not arise.
+
```
olgabot:
```suggestion
Disambiguating a residue doubles the readings of every window it falls in, so a window
containing *n* ambiguous residues yields 2^*n* disambiguated k-mers. kmerseek disambiguates a k-mer with at most 10% ambiguous residues, or up to ceil(ksize/10) ambiguous amino acids. A window with more is dropped: indexing only part of its k-mers would make matching depend on which subset was kept, which is worse than losing that one window. SwissProt holds about 900 non-canonical residues in 207.6 M, so a window with many should not arise.
```

README.md
```diff
@@ -144,43 +141,191 @@ the others) -- use it to tame proteome-scale searches where a gene can have doze
...
+chemistry per position while still collapsing the substitutions that proteins tolerate
+most often.
+
+Peterson et al. (2009) benchmarked over 150 published clustering schemes against DALI
+fold assignments and found that reduced alphabets beat the full 20-letter alphabet,
+with the best results at 9-12 classes. Ieremie et al. (2024) reused their top three and
+added five more when testing how alphabet reduction affects protein language models.
+Where the two papers overlap, their partitions are identical.
+
+| Moltype | Classes | Clusters | Source |
```
olgabot:
```suggestion
| Alphabet | Classes | Clusters | Source |
```

src/rust/hash_functions.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    /// Table-backed alphabets are applied by `encode_by_moltype` itself, since a table
+    /// cannot be expressed as a `fn(u8) -> u8`. Without that branch they would fall
+    /// through to the identity and encode to the input unchanged.
+    #[test]
+    fn test_encode_by_moltype_applies_the_reduced_table() -> Result<()> {
+        // SDM12 writes each class as its first residue in lowercase, so KER is k,
+        // TSQ is t, YF is y and LIVM is l.
+        assert_eq!(encode_by_moltype("MKTAYIAKQR", "sdm12")?, "lktaylaktk");
```
olgabot: let's rename to `encode_by_alphabet`

olgabot (commented):
Let's change moltype -> alphabet everywhere so it's not confusing

src/rust/hash_functions.rs
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Renamed in `852243f`, 29 references across `hash_functions.rs`, `sketch.rs`, `benches/benchmark.rs` and `test_hp_encoding.rs`. "Moltype" is sourmash's word for the *name* of an alphabet, so it described the argument's spelling rather than what the function does with it.

README.md
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Applied to the code in `e673b69`, not just the README — the cap was still a flat `MAX_AMBIGUITY_CODES_PER_WINDOW: usize = 4`. It is now `max_ambiguity_codes(ksize) = ksize.div_ceil(10)`, and `disambiguate_kmer` reads it off the window's own length, so no signature changed.

You were right that the fixed number was measuring the wrong thing: four codes is 80% of a k=5 window and 8% of a k=50 one, so it was tied to the reading count it produced rather than to how ambiguous the window was. Rounding up keeps a single code legal at every k-size.

Two expectations moved with it:

- `ACDEFXBZJ` at k=5 goes from 16 k-mers to 4. The cap there is one code, so the windows `EFXBZ` (two codes) and `FXBZJ` (three) are dropped.
- The disambiguation tests used windows too short to hold more than one code under the new rule (`BZ`, `BJZ`, `BBBB`). They now cut windows from human BCL-2 residues 1-30 (P10415), writing Asp10 as `B`, Glu13 as `Z` and Ile14 as `J`, so the real fragment is one of the readings that comes back rather than the tests checking invented letters against themselves.

README.md
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Done. The section is gone from the README, and I also removed the sentence it left stranded ("No hash function changed, so rebuilding under the current name gives the same hits."), which had ended up under sourmash compatibility where it no longer had an antecedent.

It survives in the PR description, where a note about which indexes stop opening is still worth having for anyone with an index on disk.

README.md
olgabot: > 🤖 _Reply written by Claude Code, posted from olgabot._

Applied — the header reads `| Alphabet | Classes | Clusters | Source |`. The table has since been folded into the single `## Alphabets` section in `1ed7d68`, which carries the same header.

PR #48, opened 2026-09-10
Store the search cache in chunks so a large index can be written at all

olgabot:
Olga: We NEED to be able to index UniRef100 ([221,143,266 sequences](https://www.uniprot.org/uniref/)) to be competitive with Foldseek/Folddisco/MMseqs2. Currently, Kmerseek can't even index SwissProt ([575,748 sequences](https://web.expasy.org/docs/relnotes/relstat.html)), which is ~400x smaller. This is a major problem, so we need to be able to index SwissProt and then see what we need to do to index UniRef100. RocksDB can only save  4GB at a time, so this PR saves larger databases in 4GB chunks.

---
## The bug

Indexing reviewed Swiss-Prot minus Ascidiacea (483_966 targets, 115_749_594 unique k-mers) serialized the `SearchCache` to 4_434_640_058 bytes and then died:

```
[save] SearchCache serialized to 4434640058 bytes in 5.1s, writing to RocksDB...
Error: Database(Error { message: "Invalid argument: value is too large" })
```

RocksDB stores a value's length in a `u32`, so no single value can reach 4 GiB. `save_inverted_index` put the whole cache under one `search_cache` key:

```rust
let serialized = bincode::serialize(&cache)?;
self.db.put(b"search_cache", serialized)?;
```

That capped every kmerseek index at ~4 GB of serialized cache no matter how much memory the machine had. It is a format limit, not a resource one, so nothing about the node, the allocation or the walltime could have got past it. This is the first reference here large enough to hit it. A 2-letter alphabet at k=23 has at most 2^23 possible k-mers, so its cache stays small; only the `protein20 k10` arm failed.

## The fix

Bytes are split across `search_cache_chunk_{i}` at 1 GiB each, with the count in `search_cache_chunks`. A cache that fits keeps the single `search_cache` key and its byte-for-byte layout, so this only changes databases that could not previously be written. Every existing index still loads.

Write order:

1. Delete `search_cache_chunks`, then `search_cache`.
2. Write each chunk.
3. Delete leftover chunks past the new count.
4. Write `search_cache_chunks` last.

Deleting the count first keeps a crash partway through from leaving a valid count standing over a mixture of old and new chunks, which would deserialize into a plausible but wrong index. Writing it last makes its presence certify that every chunk is on disk.

On read, chunks present with no count key returns `CorruptIndex` rather than `Ok(None)`. Chunk 0 is written before any other, so its presence without a count means the write did not finish. `Ok(None)` is what a database written before the cache existed returns, and the caller answers it by loading every signature and searching on, which would hide an incomplete index behind a slow search. A database that has no cache at all still returns `Ok(None)`; one test asserts both.

`CorruptIndex` is a new error variant for that state: present and broken, as distinct from absent (`NoSavedState`) or RocksDB itself refusing (`Database`).

Step 3 clears stale chunks from a larger previous cache. They are never read, since the count bounds the loop, but they are indistinguishable from interrupted-write debris and would make a healthy database fail the torn-write check on its next rewrite.

## Tests

The trigger is 4 GiB, so `write_search_cache_chunked` takes the chunk size as a parameter and the tests drive the same path at 1 KB, costing milliseconds:

- round trip across several chunks returns all three fields unchanged, checking byte order across boundaries, since a reversed or skipped chunk still deserializes into something for many payloads
- a small cache still writes the single key and no count
- a deleted chunk fails with an error naming it
- a torn write errors while a database with no cache returns `Ok(None)`
- rewriting smaller clears the larger cache's extra chunks

Full suite: 230 passed, 0 failed.

## Notes for the reviewer

- Reassembly holds the serialized `Vec<u8>` alongside the deserialized structure, so peak memory roughly doubles during load, 4.4 GB extra at this scale. Fine on the 64 GB nodes this runs on, but it is a real cost, and an alternative would be deserializing a streaming reader over the chunks.
- 1 GiB is arbitrary within the safe range. It keeps the chunk count small (a 100 GB cache is 100 keys) and leaves wide margin under the limit.
- `index_metadata` is still written as one value, and it carries `combined_mins`: every unique k-mer in the database at 8 bytes each. That key hits the same limit at ~537 million unique k-mers. This index has 115_749_594, so it is under by 4.6x, but the ceiling is not far away and this PR does not move it.
- Building locally on macOS needs `SDKROOT` set or librocksdb-sys cannot find `stdint.h`: `export SDKROOT="$(xcrun --show-sdk-path)"`.
- Nothing on the cluster picks this up until a new image is built. `params.kmerseek_image` is shared with the QfO region benchmark, so the rebuild touches runs beyond the one that hit this.

Unblocks the `protein20` baseline arm in [seanome/2024-kmerseek-analysis#30](https://github.com/seanome/2024-kmerseek-analysis/pull/30).


src/rust/index.rs
```diff
@@ -510,12 +510,89 @@ impl ProteomeIndex {
...
+        let chunks: Vec<&[u8]> = serialized.chunks(chunk_size).collect();
+        eprintln!(
+            "[save] cache is {} bytes, above the {} byte RocksDB value limit -- writing {} chunks",
+            serialized.len(),
+            chunk_size,
+            chunks.len()
+        );
+
+        // Order matters, and it is chosen so that an interrupted write leaves a database
+        // that REFUSES to open rather than one that opens wrong.
```
olgabot, line 562: This feels like a race condition waiting to happen. Let's avoid

olgabot (commented):
> [!NOTE]

## The race-condition comment on `write_search_cache_chunked`

The review filed the `WriteBatch` fix under "Consider". That was the wrong section. The ordered-put sequence is correct under a crash, but correctness by ordering is fragile: adding a key, reordering a line, or adding an early return breaks the guarantee without any compiler or test noticing, and the comment block is the only thing holding it together. `WriteBatch` moves to Should fix.

What the race is, since it shapes the fix: there is no race inside one process. `new()` is the only writer and holds the `LOCK`. The two exposures are a crash mid-sequence and, after #53, a concurrent reader. Every `kmerseek search` open is now `open_for_read_only`, which reads whatever the WAL had committed at open time, so a search started while `kmerseek index` is rewriting the same path can observe the middle of the put sequence. A `WriteBatch` closes both. RocksDB commits the batch as one WAL record, so a reader or a recovering process sees the whole old cache or the whole new one.

Checked against `rocksdb-0.23.0/src/write_batch.rs` (`WriteBatch::new`, `put`, `delete`) and `db.rs:851` (`DB::write`):

```rust
// Probe for chunks left by a previous, larger cache before building the batch;
// the batch must not read the database it is about to replace.
let mut stale = chunks.len();
while self.db.get_pinned(Self::search_cache_chunk_key(stale))?.is_some() {
    stale += 1;
}

let mut batch = WriteBatch::new();
batch.delete(b"search_cache");
for (i, chunk) in chunks.iter().enumerate() {
    batch.put(Self::search_cache_chunk_key(i), chunk);
}
for i in chunks.len()..stale {
    batch.delete(Self::search_cache_chunk_key(i));
}
batch.put(b"search_cache_chunks", chunks.len().to_string().as_bytes());
self.db.write(batch)?;
```

The single-key branch should take the same shape (delete the count and stale chunks, put `search_cache`) so shrinking back under 1 GiB also stops leaking chunks. That covers the orphaned-chunks finding from the review and from the 2026-09-11 scaling comment above. The "Order matters" comment block is deleted. Keep the read-side `CorruptIndex` check and its test as defence against external damage, and rewrite its comment to say so, since this code can no longer produce "chunks without a count".

Cost: `WriteBatch::put` copies into the batch's buffer, so write-time peak memory rises by the serialized size, 4.4 GB here. The vendored RocksDB caps individual values at `u32::MAX` and does not cap the batch total. Nobody has pushed a multi-GB batch through the WAL and memtable yet, so run one real index build before trusting it at scale.

## Does #53 change the review?

The substance, no. The write path (`new()` -> `DB::open`, the only writer) is untouched by #53, and every read-side suggestion (`get_pinned`, `reserve_exact`, streaming `deserialize_from`) works on a read-only handle. Two things did change:

1. **The branch conflicts with main** (`src/rust/index.rs`, `mergeable: CONFLICTING`). The cause is the two cherry-picks on the branch, `93042d4` and `2156408`. #53 landed in a form that rewrote `load` wholesale (`af96dc4`, "Build load and get_index_parameters on open_for_search") and supersedes both. Rebase onto `origin/main` and drop those two commits. The remaining conflict is `load`'s body, where main's three-line version wins. The dark-set image built from this branch already has both fixes, so nothing on the cluster changes.

2. **The race is now reachable from the CLI.** Before #53, `get_index_parameters` took the `LOCK` before every search, so a search against an index mid-rewrite failed on the lock. Now it opens and reads. That is the right behaviour for the pipelines, and it is why atomic cache writes stop being optional.

The rest of the review stands: the log line in `write_search_cache_chunked` that prints 1 GiB as the RocksDB value limit, the stale `save_kmer_stats_only` doc comment, and the `Vec` doubling on reassembly in `load_search_cache`. The scaling comment above sizes that transient at ~410 GB for UniRef50 with protein20, so `reserve_exact` is a stopgap and the streaming reader is the fix.

PR #49, opened 2026-09-10
Disambiguate ALL alphabets

olgabot:
Olga: While the number of B/Z residues make up <0.01% of UniRef50, it's still 12,409 proteins that are totally ignored. Similarly, AlphFold ignores ALL nonstandard amino acids, including [Pyrrolysine](https://en.wikipedia.org/wiki/Pyrrolysine) (O) and [Selenocysteine](https://en.wikipedia.org/wiki/Selenocysteine) (U), which is 249,823 UniRef50 representatives standing for almost 1 million member proteins! Again, not even 1% of UniRef50, but still worth including in my opinion!

<h3 dir="ltr">UniRef50 2026_03, full scan</h3><p dir="ltr">38_840_027 clusters, 244_028_969 member proteins.</p><div dir="ltr"><div>

if we ignored... | clusters | member proteins | share of UniRef50
-- | -- | -- | --
ambiguous residues (B/J/Z) | 1_982 | 12_409 | 0.0051%
— of those, reps with ≥2 ambiguous residues, which the old cap dropped at k=10 | 343 | 1_898 | 0.0009%
everything AlphaFold skips (B/J/Z/X/U/O) | 249_823 | 939_826 | 0.64% / 0.39%

Interestingly, even *representative* sequences in [UniRef50/90/100 ](https://www.uniprot.org/help/uniref) have ambiguous codes! Which means that these sequences are completely unrepresented in structural databases like AlphaFoldDB, as their codes are ignored.

<p dir="ltr">Distribution of ambiguous residues in the representative:</p><div dir="ltr"><div>

ambiguous | clusters | proteins
-- | -- | --
1 | 1_639 | 10_511
2 | 239 | 1_040
3 | 64 | 252
4 | 15 | 398
5–9 | 18 | 40
10–22 | 6 | 167

</div></div><p dir="ltr">Cumulatively: 343 clusters with ≥2, 104 with ≥3, 25 with ≥5, 6 with ≥10.</p>

This PR expands ALL ambiguous `B`/`J`/`Z` residues and adds two k-mers for each. e.g. for `J` -> add both k-mers with `L` (Leucine) and `I` (isoleucine) in all positions. The worst it can get is 22 ambiguous residues in a protein, which is a lot but still countably many.

This affects 9/19 alphabets, almost all the alphabets with >3 classes:

<div dir="ltr"><div>
alphabet | size | B (D vs N) | J (I vs L) | Z (E vs Q)
-- | -- | -- | -- | --
polarity4 | 4 | ≠ | = | ≠
wwmj5 | 5 | ≠ | = | ≠
funcgroups8 | 8 | ≠ | = | ≠
sdm12 | 12 | ≠ | = | ≠
mmseqs12 | 12 | = | ≠ | =
wass14 | 14 | ≠ | ≠ | ≠
hsdm17 | 17 | ≠ | = | ≠
uniprot18 | 18 | ≠ | ≠ | ≠
protein20 | 20 | ≠ | ≠ | ≠

</div></div><p dir="ltr"></p>

---
The member-protein column counts every protein in a cluster whose *representative* carries the
residue. The other members are ordinary proteins that AlphaFold does model, so 939_826 is the
sequence space unreachable through the representative, not a count of proteins AlphaFold skips.
That count, for this analysis, is the 249_823 representatives themselves.

## Expand every ambiguous residue in a k-mer

`B`, `J` and `Z` each stand for two residues, and kmerseek indexes a k-mer covering one under
both readings rather than picking one. A k-mer covering *n* of them is now indexed under all 2^*n*
readings. The cap was `ceil(ksize / 10)` of them and k-mers past it were dropped, so at the default
`--ksize 10` a k-mer holding two contributed nothing.

A dropped k-mer is a hole in the index rather than a saving. The sequences it drops are the ones
that need the index most: a protein sequenced by Edman degradation or acid hydrolysis carries runs
of B and Z, has no AlphaFold model, and so has no structural route to a homolog either. Indexing
all 2^*n* readings keeps those stretches searchable.

X, U and O are unchanged. X passes through as itself, and U and O still take their canonical
analogue (C and K) under a reduced alphabet.

### What it costs

The worst real case is [sp|P00659|RNAS1_DAMKO](https://www.uniprot.org/uniprotkb/P00659/entry),
a pancreatic RNase from [Damaliscus korrigum](https://en.wikipedia.org/wiki/Korrigum), with 22 Bs
and Zs:

```
>sp|P00659|RNAS1_DAMKO Ribonuclease pancreatic OS=Damaliscus korrigum OX=9930 GN=RNASE1 PE=1 SV=1
KESAAAKFZRZHMBSSTSSASSSBYCBZMMKSRNLTQDRCKPVBTFVHZSLABVZAVCSZ
KBVACKBGZTBCYZSYSTMSITBCRZTGSSKYPBCAYKTTQAKKHIIVACZGBPYVPVHF
BASV
```

2^22 is the wrong bound. The 22 are spread across 124 positions, so no window of 30 or fewer
holds more than 9 of them:

| k | most ambiguous residues in one window | readings from that window | readings for the whole sequence |
|---:|---:|---:|---:|
| 10 | 4 | 16 | 541 |
| 16 | 6 | 64 | 1_241 |
| 20 | 7 | 128 | 2_352 |
| 25 | 8 | 256 | 5_092 |
| 30 | 9 | 512 | 10_360 |

Sequences this dense are rare: 6 UniRef50 representatives in 38.8 M hold 10 or more.

## Verification

- The expansion counts above were computed from the P00659 sequence directly, as a check on what
  the index should produce.
- `ACDEFXBZJ` at k=5 yields 16 k-mers: the five windows hold 0, 0, 1, 2 and 3 of them, so
  1 + 1 + 2 + 4 + 8. It was 4 under the old cap, which dropped the last three windows.
- Searching CED-9 against the 25-sequence BCL-2 fixture is unchanged, since that fixture carries
  no ambiguous residues.
- 233 tests pass, `cargo fmt --check` clean, zero clippy warnings.

## Relationship to #46

#46 carried this description but merged the earlier code, which allowed at most one ambiguous
residue per k-mer and dropped any window holding two. This is the implementation the description
argues for. It also renames the terminology: B, J and Z are ambiguous residues throughout, in
prose and in identifiers, rather than "ambiguity codes".

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,11 +255,11 @@ Total matches: 21
 
 ### Amino acid disambiguation
 
-Three one-letter codes stand for a pair of residues rather than a single one, because the
+Three letters stand for a pair of residues rather than a single one, because the
 method that produced the sequence could not tell the pair apart. Asn and Gln deamidate to
 Asp and Glu during acid hydrolysis, and Ile and Leu have the same mass:
 
-| code | name | stands for |
+| letter | name | stands for |
 |---|---|---|
 | `B` | Asx | `D` (Asp, aspartate) or `N` (Asn, asparagine) |
 | `J` | Xle | `I` (Ile, isoleucine) or `L` (Leu, leucine) |
@@ -306,15 +306,16 @@ Where each k-mer count comes from:
   `hpphp`, `NTAND` and `NBMES` are both `pphpp`. Disambiguating adds none, so it stays
   at 14.
 
-A k-mer may carry at most one ambiguous residue, so disambiguating turns it into two
-k-mers and no more. A k-mer carrying a second is dropped rather than indexed under part of
-its readings, which would make matching depend on which subset was kept.
+Every ambiguous residue in a k-mer expands, so a k-mer carrying *n* of them becomes 2^*n*
+k-mers. Indexing every reading rather than a chosen subset keeps matching from depending on
+which reading was kept.
 
-Capping at one rather than allowing *n* codes and 2^*n* k-mers keeps the index from growing
-on the sequences that inform it least. A k-mer covering several codes comes from a stretch
-the source method could barely read, so most of its readings are guesses about a region
-that was never determined. SwissProt holds about 900 non-canonical residues in 207.6 M, so
-two in one k-mer should be rare.
+The expansion is affordable because ambiguous residues are rare and stay sparse within any one
+window. Swiss-Prot 2026_03 holds 525 of them, 276 `B` and 249 `Z` with no `J` anywhere,
+across 146 of its 575_748 sequences. The densest window in the database holds 9, so the
+worst single k-mer expands to 512 readings, and expansion grows the index by 0.0012% at
+k=4 and 0.034% at k=30. A k-mer carrying more than 16 is dropped, which bounds memory on
+pathological input and which no real sequence approaches.
 
 `U` (Sec, selenocysteine) and `O` (Pyl, pyrrolysine) are handled differently. They are
 specific residues rather than ambiguities, so each takes its closest canonical analogue,
--- a/src/rust/aminoacid.rs
+++ b/src/rust/aminoacid.rs
@@ -9,11 +9,11 @@ pub const STANDARD_AA: [char; 20] = [
     'Y',
 ];
 
-/// Codes that carry no residue identity and so can never be reduced: X (any residue) and
+/// Amino acids that carry no residue identity and so can never be reduced: X (any residue) and
 /// the stop codon.
 pub const SPECIAL_AA: [char; 2] = ['X', '*'];
 
-/// The two residues each ambiguity code stands for.
+/// The two residues each ambiguous residue stands for.
 ///
 ///   - `B` (Asx) is `D` (Asp, aspartate) or `N` (Asn, asparagine)
 ///   - `J` (Xle) is `I` (Ile, isoleucine) or `L` (Leu, leucine)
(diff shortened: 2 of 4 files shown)


src/rust/aminoacid.rs
```diff
@@ -9,11 +9,11 @@ pub const STANDARD_AA: [char; 20] = [
     'Y',
 ];
 
-/// Codes that carry no residue identity and so can never be reduced: X (any residue) and
+/// Amino acids that carry no residue identity and so can never be reduced: X (any residue) and
```
olgabot, line 12:
```suggestion
/// Ambiguous amino acids that carry no residue identity and so can never be reduced: X (any residue) and
```

src/rust/aminoacid.rs
```diff
@@ -22,71 +22,91 @@ pub const SPECIAL_AA: [char; 2] = ['X', '*'];
...
-/// code are also the ones least worth having: a k-mer covering several codes comes from a
-/// stretch the source method could barely read, so its readings are mostly guesses about a
-/// region that was never determined. Dropping it whole rather than indexing part of its
-/// readings keeps matching from depending on which subset was kept, and SwissProt holds
-/// roughly 900 non-canonical residues in 207.6 M, so two in one k-mer should be rare.
+/// This bounds memory; it expresses no view about which readings are worth keeping. `2^n`
+/// readings would otherwise grow without bound on pathological input, such as a long run of
+/// `B`. The densest window in Swiss-Prot 2026_03 holds 9 of them, so no real sequence comes
+/// near this.
+pub const MAX_AMBIGUOUS_RESIDUES_PER_KMER: usize = 16;
```
olgabot, line 53: I thought we were removing this? B/J/Z can't get that bad. Maybe set it to 10 to prevent weird behavior? Give me also the densest window in UnIRef50/90/100.

src/rust/aminoacid.rs
```diff
@@ -139,7 +159,7 @@ impl AminoAcidAmbiguity {
     /// discarded as unknown. Under `protein20` nothing is substituted, since there is no
     /// class to fall into and rewriting U as C would assert a residue the source never had.
     ///
-    /// The ambiguity codes B, J and Z are never resolved here. They stay in the sequence and
+    /// The ambiguous residues B, J and Z are never resolved here. They stay in the sequence and
```
olgabot, line 162: This doesn'tmake sense. I thought we WERE resolving??

README.md
```diff
@@ -255,11 +255,11 @@ Total matches: 21
 
 ### Amino acid disambiguation
 
-Three one-letter codes stand for a pair of residues rather than a single one, because the
+Three letters stand for a pair of residues rather than a single one, because the
```
olgabot, line 258:
```suggestion
Three letters stand for a pair of amino acids rather than a single one, because the
```

src/rust/aminoacid.rs
olgabot: Maybe max of 10 is good enough to cover those cases. That would expand the number of kmers representing a region to 1024 which is a lot, but not millions.

README.md
olgabot: > [!NOTE]

Applied in 280b0c6.

src/rust/aminoacid.rs
olgabot: > [!NOTE]

"Resolved" was the wrong verb; it collides with the function name. `validate_and_resolve` substitutes U→C and O→K. B, J and Z are not substituted, because each stands for two residues and picking one would assert a residue the source never claimed. They pass through unchanged and `disambiguate_kmer` expands every k-mer covering one into both readings at sketch time. Reworded in 280b0c6.

src/rust/aminoacid.rs
olgabot: > [!NOTE]

Set to 10 in 280b0c6. Two properties fall out of that value:

- At the default `--ksize 10` the cap can never fire: a 10-mer holds at most 10 ambiguous residues, and that k-mer expands to 1024 readings.
- For every k up to 30, nothing in Swiss-Prot 2026_03 or among UniRef50 2026_03 representatives is dropped. Both densest windows hold 9.

UniRef50 scan (`uniref50.fasta.gz` dated 2026-09-09; 1_982 representatives carry B/J/Z, 2_572 residues in total):

| k | densest window | representative | length | ambiguous in protein | readings for whole protein |
|---:|---:|---|---:|---:|---:|
| 10 | 5 | Q7M4G8 | 415 | 12 | 779 |
| 16 | 7 | P01658 | 132 | 14 | 1_148 |
| 20 | 8 | P01658 | 132 | 14 | 2_098 |
| 25 | 9 | P01658 | 132 | 14 | 3_889 |
| 30 | 9 | P00659 | 124 | 22 | 10_360 |

Densest 30-window across the 1_982 representatives: 1 → 1_865, 2 → 74, 3 → 11, 4 → 3, 5 → 1, 6 → 1, 8 → 3, 9 → 2. A cap of 8 would drop windows in two representatives at k=30 (P01658 and P00659); 9 and 10 drop none.

Swiss-Prot gives the same picture: 146 sequences, 525 residues, densest window 9 (P01658 at k≥25). The Swiss-Prot sequence with the most ambiguous residues overall is P02080 (HBBN_AMMLE, 26 of them), more than P00659's 22, but they are spread out and its densest 10-window holds 5.

UniRef90 and UniRef100 are not on this machine (only UniRef50 is; UniRef90 is ~30 GB and UniRef100 ~90 GB compressed). The scan is a 60-line Python script that takes about 7 minutes on UniRef50 and can run on Sherlock if those files are there.

PR #50, opened 2026-09-11
Restore --scaled, with region detection that works on a sampled sketch

olgabot:
Olga: We currently can't index large databases like [UniRef](https://www.uniprot.org/help/uniref), which is essential to be competitive with FoldSeek, Folddisco, MMSeqs2, etc. Complementary https://github.com/seanome/kmerseek/pull/48, this PR reduces the size of the search database by enabling sub-sampling of k-mers, up to --scaled=10. Maybe we could increase scaled to 100? But for now, I think 1/10th the size is pretty good already, especially since our ksizes tend to range from 10-30. So by chance, we'd probably get enough coverage of a region.

I had removed `--scaled` in https://github.com/seanome/kmerseek/pull/30 because I was getting too confused about how to stitch the k-mers together to print the matched regions.

---
## Why

PR #30 removed `--scaled` because every run used 1. Nobody had measured what it does. At k=10 on a 55,486-sequence UniRef50 sample:

| `--scaled` | indexing RSS per residue | index on disk per residue |
|---|---|---|
| 1 | 683 B | 79 B |
| 2 | 362 B | 40 B |
| 5 | 162 B | 17 B |
| 10 | 87 B | 9.7 B |

Fitting `a + b/s` gives about 21 + 662/s, so roughly 97% of indexing memory is per-k-mer and sampling removes it almost linearly. Swiss-Prot at k=10 needs ~140 GB to index at scaled=1 and ~18 GB at scaled=10. It is the cheapest lever we have for the [UniRef sizes in #48](https://github.com/seanome/kmerseek/pull/48#issuecomment-5633897396).

## Why the flag alone was not enough

`find_matched_regions` only extended a region when the next shared k-mer started exactly one residue later on both sequences. FracMinHash keeps k-mers by hash value, so at scaled=s the kept k-mers on a diagonal sit about s residues apart and are almost never adjacent. Restoring the flag by itself would have turned every match into a scatter of single k-mers, with `region_n_shared_kmers` still computed as `length - ksize + 1` and the region Poisson test inflated by about s×. Nothing would have errored.

## What changed

**The dense path is untouched.** At scaled=1 the existing loop runs as before; the 223 tests on `main` pass unchanged, including every exact-value region and CSV assertion.

**A sampled sketch takes a new path.** Each shared k-mer is a seed. Its diagonal is walked outward along the stored encoded sequences while the residues agree, which recovers the full exact match the seed sits in, including the k-mers the sample dropped. A seed inside a run already found is skipped. Every region reported is a maximal exact match of at least `ksize` residues holding at least one kept k-mer. That is the dense path's set minus the matches the sample missed entirely.

CED9 vs BCL2 regions at scaled 1, 2, 5, 10

CED9 vs BCL2 at hp k=12: 13 regions dense, then 7, 4 and 1 at scaled 2, 5 and 10. Each survivor is span-identical to a dense region. A region with n k-mers survives with probability 1 − (1 − 1/s)^n; the expected survivor count at scaled=5 for these 13 is 4.3 and the observed is 4.

**`MatchedRegion` gains `n_shared`.** The dense path sets it to its consecutive count, which is `length - ksize + 1` as before. The sampled path counts kept k-mers inside the run. The region Poisson test now reads this field, and its expectation is summed over `position_prefix`, which only holds kept query k-mers, so observed and expected stay on the same footing.

**Search takes no `--scaled`.** It reads the index's value and builds query sketches with it, so the two sides' hash cutoffs cannot disagree. The old error for scaled≠1 is gone.

**`Scaled`'s doc comment was wrong.** It described positional striding ("every 2nd k-mer"). It is a hash-fraction cutoff. Fixed. The cap of 10 stays.

## Tests

- `test_sampled_regions_are_whole_dense_regions` (scaled 2, 5, 10): every sampled region matches a dense region on span, sequences and length; `n_shared` equals an independent recount of kept shared k-mers on that diagonal; count is exactly 7 / 4 / 1.
- `test_sampled_regions_scaled_5_exact`: the four survivors as literals, with `n_shared` 1, 1, 1, 2.
- `test_sampled_landmark_k15`: the 19-residue landmark at k=15 is reported at every scaled with its full span; `n_shared` 5 / 5 / 3 / 1.
- `test_exact_run_stops_at_mismatch`: the seed window must itself agree, and a substitution splits a run rather than being bridged.
- `test_sampled_regions_need_stored_sequences`: no encoded sequence, no regions, not a scatter.
- `test_cli_index_scaled_5_search_regions`: end to end through the CLI and CSV; `scaled` column is 5, `n_intersecting_hashes` is 5, the four regions match the unit test.
- `test_cli_index_rejects_bad_scaled`: 0 and 11 fail before any database exists.
- The scaled=1 `n_shared` cross-check test now asserts the field against both the independent count and the formula.

235 passed, 0 failed. clippy clean.

## On the default

The default stays 1. Nothing changes for anyone who does not pass the flag. Whether a different default (or `floor(k/2)`) is right is a separate decision; the README section says how to choose it from the shortest match you need rather than from k.

## Not in this PR

- `CHANGELOG.md` lives on the version-bump branch, not `main`; an entry goes there.
- The dense path has a latent issue with repeated k-mers: a query position that maps to two target positions interleaves in the `(qpos, tpos)` sort and can break a run early. The diagonal grouping in the sampled path does not have it. Unifying the two would fix it but changes scaled=1 output on repeat-containing sequences, so it is left for its own PR.
- `scaled` does not change selectivity: postings per k-mer is P/U and both scale by 1/s.


src/rust/tests/test_cli.rs
```diff
@@ -539,3 +539,106 @@ fn test_cli_search_bcl2_ced9() -> Result<(), Box<dyn std::error::Error>> {
...
+    assert_eq!(
+        bcl2_regions,
+        [
+            (145, 159, 130, 144, 1, "FSLYQDVVRTVGNA".to_string()),
+            (162, 181, 138, 157, 1, "QCPMSYGRLIGLISFGGFV".to_string()),
+            (253, 266, 80, 93, 1, "MIGAGVTAGAIGI".to_string()),
+            (267, 280, 200, 213, 2, "GVVVCGRMMFSLK".to_string()),
+        ]
+    );
+    Ok(())
```
olgabot, line 619: I also want to see that the full BCL2/Ced9 region is still detected

```
Ced9 pr: …RTVGNAQTD**QCPMSYGRLIGLISFGGFV**AAKMMESVE…
Ced9 hp: …pphhphppp**pphhphhphhhhhphhhhh**hhphhpphp…
                   |||||||||||||||||||
BCL2 hp: …hhphhpphh**pphhphhphhhhhphhhhh**phpphppph…
BCL2 pr: …FATVVEELF**RDGVNWGRIVAFFEFGGVM**CVESVNREM…

- RDG starts at 138-157
```

olgabot (commented):
Update changelog too

olgabot (commented):
> [!NOTE]

## Verdict

The diagonal-walk design is sound and the `n_shared`/`position_prefix` bookkeeping holds up under inspection. The one blocking bug was that `--scaled > 1` with the default `protein20` alphabet produced no regions at all (empty CSV, matches still counted on stderr), because the sampled path required an encoded sequence that `protein20` sketches never store. That and the two should-fix items are pushed to this branch as 747bb23, 9488361 and 88d9d07; the review below is of the branch as it was at 018cb6a, with the fix noted under each item.

## Tooling

Checked out at 018cb6a, rustc 1.89.0, MSRV 1.82, edition 2021.

- `cargo fmt --check`: clean.
- `cargo clippy --all-targets`: clean. With `-W pedantic -W nursery` the new code only trips `cast_sign_loss`/`cast_possible_wrap` on the `usize`↔`isize` diagonal arithmetic, which matches the file's existing style, plus `missing_panics_doc` on `find_matched_regions`.
- `cargo test -- --test-threads=1`: 235 lib + 9 CLI passed at 018cb6a; 248 + 9 after the fixes. (`--test-threads=2` SIGABRTs on RocksDB lock contention, the known flake.)
- Merges cleanly with `origin/main` after #53.
- `bcl2.fasta[138..157] == "RDGVNWGRIVAFFEFGGVM"`: the `test_exact_run_stops_at_mismatch` fixture is real sequence.

## Blocking

**`src/rust/search.rs:1480` (at 018cb6a) — `find_sampled_regions` returned nothing for `protein20`, the default alphabet.** Fixed in 747bb23.

`sketch.rs:553-556` skips `set_encoded_sequence` when the alphabet is `protein20` ("the full alphabet encodes to itself"), so `get_moltype_sequence()` is `None` and the `let-else` bailed. The dense path tolerates this (it emits regions with `moltype_seq: String::new()`); the sampled path did not. Reproduced at unit level with BCL2 against an identical copy of itself at k=10:

```
protein20      scaled=1 shared_hashes=230 has_moltype_seq=false regions=1
protein20      scaled=2 shared_hashes=117 has_moltype_seq=false regions=0
protein20      scaled=5 shared_hashes= 44 has_moltype_seq=false regions=0
hp_lehninger2  scaled=1 shared_hashes=205 has_moltype_seq=true  regions=150
hp_lehninger2  scaled=2 shared_hashes=105 has_moltype_seq=true  regions=13
```

and through the CLI, the 25-sequence test FASTA against its own index with every filter open:

```
protein20     scaled=1: Total matches: 4   csv_rows=8
protein20     scaled=5: Total matches: 4   csv_rows=0
hp_lehninger2 scaled=1: Total matches: 596 csv_rows=55447
hp_lehninger2 scaled=5: Total matches: 600 csv_rows=13038
```

The fix walks the raw sequence when no encoded copy is stored (the idiom `index.rs:829` already uses) and leaves `moltype_seq` empty in that case, matching the dense path's CSV. Case is already normalized on both production paths (`index.rs:2028`, `main.rs:531`), so a byte compare on the raw sequence is fine. `test_sampled_regions_protein20_fall_back_to_raw_sequence` covers it with BCL2 vs BCL-xL, which share the 16-residue BH1 stretch `ELFRDGVNWGRIVAFF` (7 k-mers at k=10): reported whole with `n_shared` 7, 4, 3 at scaled 1, 2, 5, and missed at 10 where none of the 7 survive.

## Should fix

**`src/rust/main.rs:494` — `--query-is-index` never checked the query index's `scaled` against the target's; a mismatch now panicked.** Fixed in 9488361.

Before this PR every index had scaled=1 by construction, so the gap was unreachable. Now `search --query-is-index a_scaled5.db --target b_scaled1.db` loaded the query sketches as-is and the first pair with a shared hash tripped the new `assert_eq!(query_sketch.scaled(), target_sketch.scaled())` with a bare panic. The fix compares `(ksize, scaled, moltype)` after loading the query index and returns a `ValidationError` naming both sides; `test_cli_query_is_index_rejects_scaled_mismatch` covers it. (ksize/alphabet had the same pre-existing gap and are covered by the same check.)

**`README.md:102` — "Indexing holds every k-mer of every sequence in memory, about 680 bytes per residue" would be false the day #51 lands, and "search-time memory" was claimed but not measured.** Reworded in 88d9d07: the paragraph now describes the per-k-mer share of the cost (from the `21 + 662/s` fit), scopes the RSS column to the indexer in this release, and drops the search-time claim.

## Consider

**`src/rust/search.rs:1461` — "the same set the dense path finds, minus any match the sample missed" was slightly stronger than what the code does.** Doc updated in 747bb23. The walk extends while *residues* agree; the dense path extends while *sketched k-mers* are consecutive. They differ wherever a k-mer was never sketched: a window `disambiguate_kmer` rejects, or a low-complexity window under `--remove-low-complexity`. The dense path breaks the run there, the walk goes through. The sampled behaviour is the more defensible one, so the code is unchanged and the doc now says so.

**`# Panics` on `find_matched_regions`.** It is `pub` with three `assert_eq!`s on caller-supplied sketches. Added in 747bb23.

**Region ordering tie-break.** `regions.sort_by_key(|r| r.start)` is stable and diagonals come out of a `BTreeMap`, so two regions with the same query start (a repeat in the target) are ordered by diagonal, deterministically. No change; noting it was checked since the PR description says the dense path has an interleaving problem with repeats and this path does not.

## Verified clean

- **`n_shared` and `region_expectation` are on the same footing.** `sketch.rs:531` only records a position in `kmer_positions` when `hashvals.contains(&hashval)`, and `hashvals` is sourmash's post-cutoff `mins()`. `build_position_prefix` walks `kmer_positions`, so at scaled=s both observed and expected count only kept k-mers.
- **`exact_run_around` bounds.** `agree` guards `i < query.len()`, `j >= 0`, `j < target.len()` before indexing; the right walk terminates at `query.len()`, the left at 0; a seed with `qpos + ksize > len` returns `None` through `all(agree)` rather than panicking.
- **`n_shared` counting is exact.** Over sorted seeds on a diagonal, any seed `p` in `[start, end-k]` before the current one has an agreeing window (it is inside the maximal run), so it would have emitted this same run and set `covered_until` past the current seed. Seed `i` is therefore the first in the run and `take_while(p <= last_start)` counts every kept k-mer in it. `recount_shared_on_diagonal` agreeing on every region is consistent with that.
- **Clone and allocation audit.** No new reflexive clones. `query_name`/`target_name`/`moltype` are cloned once per emitted region, as on the dense path. The subseq `.to_string()`s escape into `MatchedRegion`. `seeds_by_diagonal` and `seeds.sort_unstable(); dedup()` are one small allocation per (query, target) pair, and `dedup` is needed because an ambiguous residue can put two hashes at one position.
- **`shared_position_pairs`** is a pure extraction of the dense path's first half; the dense loop is otherwise identical to `main` apart from `n_shared: consecutive_count as u32`, and all 223 pre-existing tests pass unchanged.
- **`Scaled::new` in `index` fails before RocksDB is touched** (`main.rs:256-261`) and the test asserts no directory is left behind. The `Scaled` doc fix is right: `KmerMinHash::new(scaled, ...)` at `sketch.rs:265` is a hash-space cutoff, not a stride.
- **README arithmetic**: 1 − 0.8¹ = 20% and 1 − 0.8⁸ = 83% at k=12 check out.
- **No conversational artifacts** in the new code or comments.

## Also on this branch

- The landmark region asked for in the inline comment is now asserted through the CLI at every scaled in `test_cli_landmark_region_across_scaled` (88d9d07): CED9 162..181 against BCL2 138..157, `QCPMSYGRLIGLISFGGFV` / `RDGVNWGRIVAFFEFGGVM` / `pphhphhphhhhhphhhhh`. At k=12 it is reported whole with 8, 4 and 1 of its 8 k-mers at scaled 1, 2 and 5, and missed at scaled 10, where all 8 hash above the cutoff; at k=15 it is reported at every scaled with 5, 5, 3 and 1 of its 5. The k=12/scaled=10 miss is what the README's survival formula predicts for a match this short, so the test asserts it rather than hiding it.
- Changelog entry for #50 is on `olgabot/bump-version-0.4.0` (a161199), where `CHANGELOG.md` lives.

src/rust/tests/test_cli.rs
olgabot: > [!NOTE]

Added in 88d9d07 as `test_cli_landmark_region_across_scaled`, through the CLI and CSV. The row for CED9 162..181 / BCL2 138..157 is asserted with `region_subseq = QCPMSYGRLIGLISFGGFV`, `target_subseq = RDGVNWGRIVAFFEFGGVM`, `moltype_seq = pphhphhphhhhhphhhhh` and `region_length = 19`, and `region_n_shared_kmers` is asserted per case:

| k | scaled 1 | 2 | 5 | 10 |
|---|---|---|---|---|
| 12 | 8 | 4 | 1 | missed |
| 15 | 5 | 5 | 3 | 1 |

At k=12 and scaled=10 all 8 of its k-mers hash above the cutoff, so the match is genuinely gone at that setting; the test asserts the miss rather than skipping the case. At k=15 one of its 5 k-mers hashes low enough to survive scaled=10 and the whole 19 residues come back from it.

PR #53, opened 2026-09-13
Open the index read-only when autodetecting search parameters

olgabot:
Olga: I kept getting errors in my Nextflow pipelines (e.g. https://github.com/seanome/2024-kmerseek-analysis/pull/30, https://github.com/seanome/2024-kmerseek-analysis/pull/39) where it couldn't search the same Kmerseek index at the exact same time due to RocksDB LOCK files existing. But these are just for searching, so we shouldn't need to worry about concurrent reading since we're not writing to the index. This PR opens the index read-only so that the same index can be read multiple different times.

---
## What

`ProteomeIndex::get_index_parameters` (the "Autodetecting parameters from target database" step of `kmerseek search`) opened the RocksDB with `DB::open`, which takes the exclusive `LOCK` file. `open_for_search` has been read-only since 4e6c326, but this call runs before it, so two searches starting at the same moment against one index raced on the lock. Now `DB::open_for_read_only`, same as `open_for_search`; the function only reads metadata.

`load` and `get_index_parameters` now build on `open_for_search` instead of each opening the database themselves, so there is one read-only open path and the schema-version check runs on every open.

## Why

On the invertebrate dark-set pipeline (one shared index on Lustre, 23 query chunks × 4 alphabet/mask combos), 32 of 92 search tasks died with:

```
Error: Database(Error { message: "IO error: While lock file: minus_Ascidiacea.hp_thomas_dill2.k23.lcfalse.kmerseek.rocksdb/LOCK: Resource temporarily unavailable" })
```

The pipeline's `|| true` turned those into 32 empty result files, and the "fraction of dark proteins kmerseek reaches" number came out wrong and asymmetric between mask on and mask off (7 vs 3 chunks lost).

## Verification

- `cargo test --lib get_index_parameters` passes.
- 12 concurrent `kmerseek search` processes against one freshly built index: old binary 11/12 fail on the lock, this branch 12/12 succeed, in both FASTA-query and `--query-is-index` modes.


src/rust/index.rs
olgabot, line 1367: Do we really need `pub fn load` in addition to `pub fn get_index_parameters`? Why can't they be the same?

olgabot: *Written by Claude Code, posted from Olga's account.*

They return different things. `get_index_parameters` returns `(ksize, scaled, moltype)` and closes the database. `load` returns a `ProteomeIndex` with every signature deserialized into memory, which the `--query-is-index` path needs. So one can't replace the other.

But each carried its own copy of "open read-only, read metadata", and `open_for_search` had a third. This PR fixed the same lock bug in two of them separately, which is the cost of that duplication.

Collapsed both onto `open_for_search` in af96dc4:
- `load` is `open_for_search` + `load_state` (3 lines, was ~95)
- `get_index_parameters` is `open_for_search` + three field reads (2 lines, was ~50)
- the check for an index built by a newer binary moved into `open_for_search`, so every read-only open runs it, not only the autodetect step
- dropped the `index_state` fallback. Nothing has written that key since 22801df (Sep 2025), and `ProteinSearcher::load`, which runs right after autodetection, requires `index_metadata`, so search already fails on those indices.

223 lib tests pass. 12 concurrent searches on one index: 12/12 in both query modes.

src/rust/index.rs
```diff
@@ -1454,23 +1362,21 @@ impl ProteomeIndex {
...
 
-        // Validate schema version before loading anything else.
-        // Indices built before versioning was added have no schema_version key and are
-        // treated as version 0.
-        // Version 0 indexes built after commit 9d083c8 (Feb 24 2026) use kmer_positions
-        // format and are fully compatible with schema version 1. We accept them here.
-        // Only truly incompatible formats (e.g., pre-Feb-24 kmer_infos format) need rebuilding,
-        // but those can't be detected by this key alone.
-        let stored_version = Self::read_schema_version(&db)?;
+    /// Refuse an index written by a newer binary than this one.
```
olgabot: Why this refusal? I don't care about the old indices. This package doesn't have a ton of users. I care about making the future better

olgabot: *Written by Claude Code, posted from Olga's account.*

The refusal is for the future, not the past: it rejects an index with a *newer* schema than the binary. When #51 lands schema 3, a 0.4.0 binary opening a schema-3 index gets "upgrade the binary" instead of a bincode length-mismatch error. Old indices are what it accepts, and the doc comment spent three lines on that, which is the wrong emphasis. Rewritten in 3ea4609 to say only what the check protects against.

The code that exists for old indices is `ProteomeIndexState` (the `index_state` fallback in `load_state`), `LegacyProteomeIndexMetadata` and the schema<2 branch in `read_metadata`, plus one test that rolls `schema_version` back. #51 touches those lines and adds a schema-2 read path of its own, so deleting them here would conflict. Follow-up once #51 merges, filed as https://github.com/seanome/kmerseek/issues/55: schema must equal the binary's (`!=`, one "rebuild the index or upgrade the binary" error), and every legacy read path goes.

PR #54, opened 2026-09-13
Extend matched regions past their exact seed, and rank them with a Karlin-Altschul E-value

olgabot:
Olga: Exact k-mer matching doesn't do well with <60% identity, but I'd like to have Kmerseek perform well in the [twilight zone](https://pubmed.ncbi.nlm.nih.gov/10195279/) (20<= x <= 35% identity). This PR allows for that by using the HP k-mers as an initial seed, then extending, like BLAST does. Though BLAST uses a small initial seed, like 3 amino acids, and this is longer.

CleanShot 2026-09-17 at 08 54 55@2x

This PR uses the k-mers as an anchor, then extends an ungapped alignment until there's too many mismatches in the alphabet (mismatch penalty: `--extend-mismatch-penalty C`), as defined by `--extend-xdrop X`. By default, `X = 8` and typically, we set the mismatch penalty to `C=2`.

The Karlin-Altschul E-value comes from BLAST:

- Karlin S, Altschul SF (1990). Methods for assessing the statistical significance of molecular sequence features by using general scoring schemes. *PNAS* 87:2264-2268. https://doi.org/10.1073/pnas.87.6.2264 (λ and K for ungapped local alignment; the E-value formula)
- Karlin S, Altschul SF (1993). Applications and statistics for multiple high-scoring segments in molecular sequences. *PNAS* 90:5873-5877. https://doi.org/10.1073/pnas.90.12.5873 (sum statistics, used by `--chain-max-gap`)
- Altschul SF, Gish W, Miller W, Myers EW, Lipman DJ (1990). Basic local alignment search tool. *J Mol Biol* 215:403-410. https://doi.org/10.1016/S0022-2836(05)80360-2 (BLAST itself: seed, ungapped extension, E-value)
- Altschul SF, Madden TL, Schäffer AA, Zhang J, Zhang Z, Miller W, Lipman DJ (1997). Gapped BLAST and PSI-BLAST: a new generation of protein database search programs. *Nucleic Acids Res* 25:3389-3402. https://doi.org/10.1093/nar/25.17.3389 (X-drop extension)
- Schäffer AA, Aravind L, Madden TL, Shavirin S, Spouge JL, Wolf YI, Koonin EV, Altschul SF (2001). Improving the accuracy of PSI-BLAST protein database searches with composition-based statistics and other refinements. *Nucleic Acids Res* 29:2994-3005. https://doi.org/10.1093/nar/29.14.2994 (λ from the pair's own composition)
- Altschul SF. The statistics of sequence similarity scores. NCBI BLAST tutorial. https://www.ncbi.nlm.nih.gov/BLAST/tutorial/Altschul-1.html (plain-language walk-through of E = K m n e^(-λS))

---

## Why

An exact run in the encoded alphabet is a seed, not the match. Between remote homologs the HP class of aligned residues agrees far above chance per position (Cohen's κ 0.46 at 20-30% identity, twice protein20), but the longest exact HP run between such pairs averages 12 residues, so almost none of them share an exact 23-mer, the seed length of the exact arm (2024-kmerseek-analysis nb 230). That is why every benchmark and every fitted constant in this PR is at k = 12: it is the longest seed that about half of the true pairs at 20-30% identity still share. The exact-run length was the limit, not the biology.

## What changed

Three pieces, each described below: the extension out of a seed, a Karlin-Altschul E-value for the extended region, and a calibration of that E-value's constants when the index is built. New flags: `--extend-mismatch-penalty`, `--extend-xdrop`, `--ka-k`, `--ka-queries`, `--ka-seed`, `--ka-null`, `--ka-reference`, `--ka-survival-out` (index only), `--chain-max-gap`, `--chain-max-shift`.

**Extension (31b510f).** `--extend-mismatch-penalty C` grows each region outward along the stored encoded sequences at +1 per agreeing position and -C per disagreeing one, stopping when the running score has fallen `--extend-xdrop X` (default 8) below its best. Seeds on one diagonal whose extensions meet merge into one region. Default off.

`MatchedRegion` gains `n_shared` (exact shared k-mers, summed on merge, never recomputed from the extended length) and `n_mismatches`. The region Poisson test reads `n_shared`, which is unchanged for exact regions. CSV gains `region_n_mismatches`.

**Karlin-Altschul score (bd120d3).** Extension alone moved region boundaries but not ranks: the Poisson test still counts only exact shared k-mers, so a region that grew across mismatches scored the same as before it grew. Each extended region now also gets the score BLAST uses for an ungapped alignment, computed in the encoded alphabet:

```
S    = matches - C × mismatches
E    = K × m × n × e^(-λS)
bits = (λS - ln K) / ln 2
```

- `S`: raw score of the region. Each position where query and target fall in the same class adds 1; each position where they differ subtracts C.
- `C`: the mismatch penalty, `--extend-mismatch-penalty`. Default 0, which turns extension and this score off. The benchmarks in this PR and the fitted K use C = 2.
- `a`: the chance that two positions drawn at random, one from the query and one from the target, land in the same class. Computed from this pair's own class frequencies (sum over classes of p_query × p_target), following Schäffer et al. 2001 rather than one database-wide value. A balanced pair in a 2-class alphabet has `a ≈ 0.5`; a pair of mostly-hydrophobic sequences has `a` near 1.
- `λ` (lambda): the scale that converts a raw score into a probability (Karlin & Altschul 1990). For independent positions it is the positive solution of `a·e^λ + (1-a)·e^(-Cλ) = 1`, solved by bisection once per pair. A positive solution exists only when a random aligned position has a negative expected score, i.e. `a < C/(1+C)` (2/3 at C = 2). Two mostly-hydrophobic sequences are above that line: agreeing is what their compositions do by chance, λ is 0, and no run of agreement is significant at any length. That is the transmembrane failure mode of the Poisson count, and it is gone by construction. Real proteins are not independent positions, so the calibration below measures a correction factor, r_database, and every pair's λ is multiplied by it (0.95 on SCOPe40, 0.85 on Swiss-Prot).
- `m`: query length in residues.
- `n`: residues in the database (the index's k-mer count stands in for it).
- `K`: the constant that turns the search space into an expected count. It depends on the alphabet, the seed length, C, X and the database, so it is fitted on the index itself when the index is built (see "Calibration" below) and stored in it: 0.023 on SCOPe40, 0.016 to 0.018 on Swiss-Prot, 0.014 on UniRef50, all at `hp_thomas_dill2` k = 12, C = 2. `--ka-k` overrides it; it has no default any more.
- `E`: the expected number of regions scoring ≥ S between an unrelated query of length m and a database of n residues. Smaller is better. `bits` is the same evidence on BLAST's bit scale, where larger is better.

lambda against a at C = 2: the root falls to 0 at a = 2/3

CSV gains `region_ka_bits` and `region_evalue`. Both are 0 / inf without the extension flag.

**Calibration at index time (76b32c9, 8b5f753, 1b79bd5).** There are two λ-like quantities in this PR and they do different jobs.

- **λ for the pair** (`λ_pair`): the closed-form λ from the two sequences' compositions, described in the bullets above. "Closed form" means it comes out of an equation, `a·e^λ + (1-a)·e^(-Cλ) = 1`, not out of data; the equation is Karlin and Altschul's, and it is exact when every position of both sequences is drawn independently from the pair's compositions. Computed by the search for every query-target pair. It says how many nats one unit of raw score is worth under that independence assumption.

  Worked example. BCL2 (human, P10415) against CED9 (worm) in `hp_thomas_dill2`: both are about half hydrophobic, so the chance two random positions share a class is a = 0.5, and the equation gives λ_pair = 0.481 nats per score unit at C = 2. A region between them with 48 agreeing positions and 4 disagreeing has S = 48 − 2 × 4 = 40 and is worth 0.481 × 40 = 19.2 nats: under independence, E = K·m·n·e^(−19.2).

- **r_database** (`ka_r_database` in the code): one number per index, measured on real sequences when the index is built. Real proteins are not independent positions (hydrophobic runs, helix and strand periodicity), so the equation's λ is off by a factor, and this is that factor. 1.00 means real sequences behave exactly as the equation assumes and no correction is needed. 0.947 (SCOPe40) means one unit of raw score is worth 5% fewer nats than the equation says: the same BCL2/CED9 region is worth 0.481 × 0.947 × 40 = 18.2 nats and its E-value is e^(19.2 − 18.2) = 2.7× larger than under independence. 0.85 (Swiss-Prot) makes the region worth 16.4 nats and its E-value 7.5× larger.

What it means for the closed form to be right. The equation assumes each residue is drawn like a weighted coin flip with no memory of the residue before it. Real proteins have memory: a hydrophobic residue is more often followed by another (membrane segments, cores), and helices and strands repeat with a period. So the question is whether real sequences produce high-scoring regions at the rate coin flips would. That is testable. Shuffle the sequences so they *are* coin flips with the same composition, search them, measure how fast the counts fall with score, and divide by the equation's λ: the answer should be 1.00. On SCOPe40 it is 1.04 for all three kinds of scrambled sequence, which is the check that the fitting code reads the right thing. Then do the same with the real, unscrambled sequences. If the answer were still 1.00 the equation would be right for real sequences too and no correction would be needed. It is 0.947 on SCOPe40 domains and 0.83 to 0.87 on full-length proteins: real sequences reach a given score 5% to 15% more easily than coin flips, because their runs and periodicity make long agreeing stretches more likely. r_database is that measured ratio.

Why two numbers and not one: λ_pair carries the composition of *this* pair (two hydrophobic proteins get λ_pair = 0 and no significance; a balanced pair gets 0.48). The correction carries what real sequences in *this database* do beyond composition, which is the same for every pair. Splitting them is the same move BLAST's composition-based statistics make: keep the per-pair composition adjustment, fit the rest once.

A search combines them as

```
E = K × m × n × e^(−λ_pair × r_database × S)
```

**K** is the constant in front, also one number per index, measured at the same time.

How they are measured:

1. Take 200 sequences from the database itself and search each one against the whole index (all 15,177 SCOPe40 domains, or all 15,000 of a sample), not against the other 199, exactly as a user's query would be searched (C = 2, X = 8, every filter off). Hits of a sequence on its own database entry are dropped.
2. A region is one gapless local alignment between a stretch of the query and a stretch of one target: an exact shared k-mer run, extended outward position by position at +1 per agreeing class and −C per disagreeing one until the X-drop stops it, merged with any other seed on the same diagonal it grows into. One query-target pair can give several regions on different diagonals, and every region from every pair counts once. For every region, compute x = λ_pair × S: the region's raw score in the pair's own λ units. x is in nats. A region at x = 15 would be as rare as e^(−15) under the independent-positions model.
3. Count the regions in bins of x half a nat wide, and plot ln(count) against x. If the independent-positions model were exactly right, the points would fall on a straight line with slope −1, and the line's height would give K.
4. Fit a straight line to the points. r_database is that line's slope with the sign flipped. K comes from the height: count in the bin at x = K × L × N × (1 − e^(−w)) × e^(−r_database × x), with L the residues in the 200 queries, N the residues in the database and w = 0.5 nat the bin width (Altschul & Gish 1996; Pearson 1998, the label-free route in Altschul's tutorial).
5. Decide which bins to fit. Among 200 real sequences some are related to sequences in the database, and related pairs produce regions with high x that unrelated pairs never would. Those bins must not be fitted. To find where they begin, the same 200 sequences are shuffled so that every pair of neighbouring residues occurs as often as in the original (this keeps the lengths of hydrophobic runs; Altschul & Erickson 1985, Kandel et al. 1996; `--ka-reference shuffled-dipeptide`), searched the same way, and counted in the same bins. Shuffled sequences have no relatives. Bin by bin, ln(real count) − ln(shuffled count) is taken; where the real sequences have only the structure the shuffle keeps, that difference is flat. The first bin where it rises more than two Poisson standard deviations above a line through its eight lowest bins is where relatives begin, and the fit stops there. The line is read off the eight highest bins below that point.

The fit goes to the count in each bin, not the count at or above it, because a plateau of related pairs far up the axis would add a constant to every summed count below it and flatten the slope there too.

Why x and not the raw score S: Swiss-Prot holds pairs of membrane and low-complexity proteins whose raw scores run past 60 while ordinary pairs stop near 45. In raw score the two populations have different slopes and one line cannot serve both. In x each pair is already on its own λ; the biased pairs have λ_pair near 0 and sit at x = 0, out of the way.

The fit (r_database, K, the bins used, where relatives begin, both curves) is stored in the index under its own RocksDB key and printed by every search that uses it. A search with a different C or X fits its own on `--ka-queries` sequences, or is refused unless `--ka-k` is given. `--ka-null` picks what the calibration queries are (`database`, the default, or one of the three scrambled kinds below); `--ka-reference` picks the reference the `database` fit is cut off against (`shuffled-dipeptide` by default); `--ka-seed` fixes which sequences are sampled.

Normalised region score distributions on SCOPe40, a Swiss-Prot sample and a UniRef50 sample, with the fitted line and the dipeptide-shuffled reference

Top row: regions in each half-nat bin of x, the points the line is fitted to; grey triangles are the same queries shuffled keeping dipeptides. Bottom row: the same regions summed. Grey band: the 8 bins used. Dotted line: 30 regions, below which a bin is not fitted. ± is the standard error of r_database. Drawn by `scripts/plot_ka_survival.py` from the curves under `docs/images/ka_fit_curves/`, which `kmerseek index --ka-survival-out` writes.

**Results by database** (`hp_thomas_dill2` k = 12, C = 2, X = 8, 200 database queries; the Swiss-Prot and UniRef50 rows are random samples, seed 1, because the index on this branch holds every signature in memory):

| database | sequences | residues | r_database | K | bins fitted (x, nats) | relatives begin at | regions counted | time, memory |
|---|---|---|---|---|---|---|---|---|
| SCOPe40 | 15,177 | 2.8 M | 0.947 ± 0.015 | 0.0227 | 13.5 to 17.5 | not reached | 14 M | 13 s, 2.8 GB |
| Swiss-Prot sample | 15,000 of 575,748 | 5.5 M | 0.865 ± 0.004 | 0.0180 | 7.5 to 11.5 | x = 11.5 | 59 M | 46 s, 11 GB |
| Swiss-Prot sample | 60,000 of 575,748 | 21.7 M | 0.851 ± 0.003 | 0.0159 | 7.0 to 11.0 | x = 11.0 | 253 M | 177 s, 41 GB |
| UniRef50 sample | 15,000 of 38.8 M | 4.7 M | 0.826 ± 0.005 | 0.0139 | 6.5 to 10.5 | x = 10.5 | 51 M | 39 s, 12 GB |

Reading the table: SCOPe40 domains behave almost like independent positions (r_database 0.95). Full-length proteins do not: their r_database is 0.83 to 0.87, so every E-value on a proteome is larger than the closed form would give. The two Swiss-Prot samples agree to 2% in r_database and 12% in K, as constants per residue should, so a sample calibrates the full database. On the proteomes related pairs begin around x = 11, which is why those fits are read lower down than the SCOPe40 fit.

What the numbers mean for a hit. For a 200-residue query and a region at x = 15, 20 and 25 nats (raw score 33, 44 and 55 when λ_pair = 0.456):

| database | E at x = 15 | E at x = 20 | E at x = 25 |
|---|---|---|---|
| SCOPe40 (n = 2.5 M) | 7.8 | 0.068 | 6.0e-4 |
| Swiss-Prot, full (n = 210 M, using the 60,000-sample fit) | 1.9e3 | 27 | 0.38 |
| old: K = 0.03 and r_database 1, SCOPe40 | 4.6 | 0.031 | 2.1e-4 |

**Every null against every database.** The rows of the grid are what the 200 calibration queries are: the database sequences as they are (with the grey dipeptide-shuffled reference the fit is censored against), then the three kinds of scrambled sequence. The columns are the databases. Each null has two rows: regions per bin (what is fitted), then the same regions summed.

Grid: four nulls by three databases

r_database by null and database (1.00 = the closed-form λ is right as it stands; below 1, sequences reach a given score more easily than independent positions would):

| calibration queries are | SCOPe40 | Swiss-Prot sample | UniRef50 sample |
|---|---|---|---|
| `database`: the sequences as they are | 0.947 | 0.865 | 0.826 |
| `shuffled-dipeptide`: shuffled, neighbouring pairs kept | 1.038 | 0.942 | 0.873 |
| `shuffled`: shuffled | 1.037 | 1.015 | 0.863 |
| `reversed`: read back to front | 1.036 | 0.858 | 0.809 |

K by null and database:

| calibration queries are | SCOPe40 | Swiss-Prot sample | UniRef50 sample |
|---|---|---|---|
| `database` | 0.0227 | 0.0180 | 0.0139 |
| `shuffled-dipeptide` | 0.0641 | 0.0288 | 0.0154 |
| `shuffled` | 0.0653 | 0.0792 | 0.0126 |
| `reversed` | 0.0704 | 0.0165 | 0.0116 |

Reading the two tables. On SCOPe40 the three scrambled nulls agree with each other and give r_database 1.04: the closed form holds for scrambled domains, which checks the fitting code. Real domains give 0.95 and a third of the K. On Swiss-Prot a plain shuffle still gives 1.0, but keeping neighbouring pairs already pulls the scale to 0.94: hydrophobic runs alone account for a third of the gap between scrambled and real. Reversed sequences on Swiss-Prot give 0.86, the same as real ones, and their curve bends upward from x = 14 (bottom-middle panels of the grid): reversed proteins still match their forward paralogs in a two-class alphabet. On UniRef50 even the shuffled curves bend upward past x = 17, from pairs whose composition is extreme without reaching the λ_pair = 0 bound; those bins lie beyond every fit window.

**Choosing the null.**

| | `database` (default) | `shuffled-dipeptide` | `shuffled` | `reversed` |
|---|---|---|---|---|
| the query is | a database sequence, unchanged | the same residues rearranged so every neighbouring pair occurs as often as before | the same residues, order shuffled | the same sequence, back to front |
| keeps | everything real: composition, hydrophobic runs, helix and strand periodicity, low-complexity stretches | composition and run lengths | composition | composition, runs, periodicity |
| loses | nothing; related pairs are cut off where the curve rises above the reference | periodicity, and any order beyond neighbours | runs and periodicity | which amino acid sits at each position (the order is reversed), but in a two-class alphabet the hydrophobic/polar pattern of a helix or strand survives reversal |
| pro | the quantity the E-value is about: hits from unrelated *real* proteins. Altschul's first option, Pearson 1998, no labels needed. Where the real curve starts to rise above the reference is a free readout of where the database's related pairs begin | keeps the hydrophobic runs that give real HP sequences their long tail (on the Swiss-Prot sample its counts past raw score 49 are 5 to 20× the plain shuffle's), so as a reference it lets only relatives and periodicity through | the model the formula was derived for; r_database should be 1 and is (1.04), which checks the fitting code. Works on a single-family database | the standard decoy for searches in the full 20-amino-acid alphabet and for HMMs; keeps real structure; no labels |
| con | costs two calibration searches. Unrelated-but-similar pairs (two membrane proteins, two coiled coils) are part of the null, which is correct for an E-value | on a family-dense database its own tail can bend upward from low-complexity pairs alone; not a null for relatives | too easy a null: real unrelated proteins have runs and periodicity that shuffled ones lack, so E-values come out too small. Altschul's own "artificiality of a random sequence model" | a helix (HPPHHPP…) or strand (HPHP…) reads nearly the same backwards, so reversed family members still seed and extend against the query one element at a time; on a 300-protein BCL2 fixture the curve bent up where `database` did, and K came out 1.8× the others. At SCOPe40's 40% identity the leak is invisible |
| use it for | the E-values | the reference the `database` fit is censored against | the check that the code reproduces the theory | a check that the leak claim holds on your data; not for fitting in an HP alphabet |

Limits, in the PR rather than in the code: one r_database and K per index describe a typical query. A query rich in membrane segments has a positive expected score against every other membrane protein; its per-pair λ is 0 (no significance) and the fit puts it at x = 0, but nothing rescues it. Masking long hydrophobic runs (the HP analogue of SEG) or a per-query fit at search time is a separate change. The Swiss-Prot and UniRef50 numbers are from samples: the full databases need main's streaming index (PR #51) merged into this branch; the sample fits agree with each other, so the full-database K should land close to them.

**Chaining (92aa969).** `--chain-max-gap G --chain-max-shift D` joins colinear extended regions at most G residues apart on the query and within D diagonals of each other into one region scored with Karlin-Altschul sum statistics (1993): each member's normalised score is lambda S_i - ln(K m n_t), and P(T_r >= t) ~ e^-t t^(r-1) / (r! (r-1)!), times the target count. `MatchedRegion.n_chained`, CSV `region_n_chained`. Default off. On SCOPe40 domains it barely moves ranking (0.066 -> 0.068 at 30/10; 2% of regions chain); its purpose is region transfer in the QfO benchmark, where a call has to cover half a domain to carry its label and one gapless run rarely does.

K refit: shuffled queries give 0.026, different-fold real pairs 0.033, reversed 0.031; the 0.03 default stands.

## What it does

On 200 SCOPe40 domains against the SCOPe40 index (details in the two comments below):

| arm | ranked by | superfamily relatives found before first different-fold hit |
|---|---|---|
| exact k=23 | Poisson | 0.0029 |
| extended k=12, penalty 2 | Poisson | 0.0012 |
| extended k=12, penalty 2 | E-value | **0.066** |
| MMseqs2, same queries | its E-value | ~0.25 |

At E <= 0.01: 109 pairs, all same-superfamily, none different-fold. The extension costs +4-6% CPU at k=12 and no memory; the 14-17 GB peak at k=12 is the search itself and is independent of this PR.

## Tests

- `test_extend_regions_bridges_one_flip`: a 25-position HP pattern with one flip is two exact seeds (5 k-mers each) and one extended region (`n_shared` 10, `n_mismatches` 1); a penalty larger than the X-drop leaves the seeds apart; penalty 0 returns them untouched.
- `test_extend_regions_contains_seeds_on_real_pair`: CED9 vs BCL2 at hp k=12; every extended region contains the seeds it grew from on its own diagonal, `n_shared` is the seeds' sum, `n_mismatches` matches a recount.
- `test_karlin_altschul_lambda`: the golden-ratio root at a = 0.5, C = 2; zero at and above the boundary; the root satisfies the equation for a 20-letter-like composition.
- `test_class_composition_match_probability`.
- `test_cli_search_extend_mismatch_penalty`: end to end through the CLI; off reproduces the 242 exact rows with `region_n_mismatches` 0, on merges regions and spans mismatches.
- `karlin_altschul::tests`: the closed-form K reduces to the textbook (q − p)²/q for ±1 scoring and to 0.1631 at a = 0.5, C = 2 (checked against a 4-million-step random walk: 458 segments simulated, 478 predicted); `fit_scores` recovers a slope of 0.4 from exact exponential counts and stops at a 400-region homolog bump without the bump reaching the bins below; `fit_scores_with_reference` stops two bins after a ramp of relatives starts and reads the slope below it; the dipeptide shuffle of human BCL2 keeps every dipeptide count, the first and last residue, and is reproducible from its seed; the sampler and the decoys are deterministic.
- `test_calibrate_ka_on_first25_is_stored_and_reused`: the same seed gives the same fit (r_database 0.870, K 0.0178, bins x = 7.5 to 11.5, 9,561 regions from 25 shuffled queries); the stored fit is found by (C, X), `--ka-k` wins over it, and a penalty without a fit and no queries allowed is refused.
- `test_cli_ka_fit_at_index_time_is_reused`: `kmerseek index --ka-queries 25` prints and stores the fit; `search` reports "stored in the index", "--ka-k", a refusal with `--ka-queries 0`, and "fitted now" for C = 3.

250 passed. clippy clean apart from the two pre-existing warnings on main.

## Not in this PR

- A two-hit rule / seed budget before extension, and streaming output. 1,000 Swiss-Prot queries at hp k=15 `--scaled 10` (with PR #50 merged in a scratch branch) took 437 s at 83 GB and reported 55.7 M pairs; DIAMOND `--sensitive` does the same queries in 17 s at 1.2 GB. The search side, not the index, is what keeps this off a laptop.
- SCOPe40 at the other alphabets and seed lengths in the benchmark (k = 23 exact arm has no extension and no fit; `hp_lehninger2`, `dayhoff6`, `protein20` at k = 12 do). The fit runs at every `kmerseek index`, so the pipeline picks them up; the numbers above are `hp_thomas_dill2` only.
- Calibration cost at proteome scale: `kmerseek index` builds a `ProteinSearcher` over the in-memory signatures for the 200 calibration searches, so a Swiss-Prot build pays the search-side memory noted below once more at the end.
- The pipeline side lives in 2024-kmerseek-analysis `olgabot/pipeline-extend-param` (`--kmerseek_extend_penalty`, `--kmerseek_search_image`, `--kmerseek_rank_by region_ka_bits`, `--kmerseek_max_evalue`) and the image `olgabot/kmerseek:2026-09-13-extend-mismatch`.

### DIFF
--- a/src/rust/index.rs
+++ b/src/rust/index.rs
@@ -510,10 +510,65 @@ impl ProteomeIndex {
             t1.elapsed().as_secs_f32()
         );
         let t2 = Instant::now();
-        self.db.put(b"search_cache", serialized)?;
+        self.write_search_cache_bytes(&serialized)?;
         eprintln!("[save] search_cache written in {:.1}s", t2.elapsed().as_secs_f32());
 
         Ok(())
+    }
+
+    /// Largest slice written under one RocksDB key.
+    ///
+    /// RocksDB refuses any single value at or above 4 GiB -- its length is a u32 -- and
+    /// returns `Invalid argument: value is too large`. That is a format limit, so no amount
+    /// of memory or tuning avoids it. A reviewed Swiss-Prot index (483_966 targets,
+    /// 115_749_594 unique k-mers) serializes to 4.43 GB and hit it.
+    ///
+    /// 1 GiB leaves a wide margin and keeps the chunk count small: even a 100 GB cache is
+    /// 100 keys.
+    const SEARCH_CACHE_CHUNK: usize = 1 << 30;
+
+    /// Store the serialized [`SearchCache`], splitting it across keys when it is too large
+    /// for one.
+    ///
+    /// A cache that fits keeps the original single `search_cache` key and byte-for-byte
+    /// layout, so databases written by this version are still readable by older builds
+    /// whenever they would have been readable at all.
+    ///
+    /// The chunk count is written LAST. A crash midway therefore leaves a database with no
+    /// count key, which reads as "no cache" rather than as a cache that silently ends
+    /// early -- the failure that would otherwise surface much later as a truncated
+    /// inverted index and quietly missing search hits.
+    fn write_search_cache_bytes(&self, serialized: &[u8]) -> IndexResult<()> {
+        self.write_search_cache_chunked(serialized, Self::SEARCH_CACHE_CHUNK)
+    }
+
+    /// The chunk size is a parameter so the split path can be tested against a few KB
+    /// instead of the 4 GiB it takes to reach it in production.
+    fn write_search_cache_chunked(&self, serialized: &[u8], chunk_size: usize) -> IndexResult<()> {
+        if serialized.len() < chunk_size {
+            self.db.put(b"search_cache", serialized)?;
+            return Ok(());
+        }
+
+        let chunks: Vec<&[u8]> = serialized.chunks(chunk_size).collect();
+        eprintln!(
+            "[save] cache is {} bytes, above the {} byte RocksDB value limit -- writing {} chunks",
+            serialized.len(),
+            chunk_size,
+            chunks.len()
+        );
+        for (i, chunk) in chunks.iter().enumerate() {
+            self.db.put(Self::search_cache_chunk_key(i), chunk)?;
+        }
+        // Any stale single-key cache from an earlier build of the same database would
+        // otherwise win on read, because the reader prefers it.
+        self.db.delete(b"search_cache")?;
+        self.db.put(b"search_cache_chunks", chunks.len().to_string().as_bytes())?;
+        Ok(())
+    }
+
+    fn search_cache_chunk_key(i: usize) -> Vec<u8> {
+        format!("search_cache_chunk_{i}").into_bytes()
     }
 
     /// Number of k-mers listed in the "most common" / "least common" summaries.
@@ -1415,13 +1470,39 @@ impl ProteomeIndex {
     ///
     /// The caller (ProteinSearcher::load) uses this to skip loading all signatures and instead
     /// find candidates via the inverted index, loading individual signatures on demand.
+    /// Reads a cache written either as one value or as chunks; see
+    /// [`Self::write_search_cache_bytes`].
     pub fn load_search_cache(&self) -> IndexResult<Option<SearchCache>> {
         if let Some(data) = self.db.get(b"search_cache")? {
             let cache: SearchCache = bincode::deserialize(&data)?;
-            Ok(Some(cache))
-        } else {
-            Ok(None)
+            return Ok(Some(cache));
         }
+
+        let Some(count) = self.db.get(b"search_cache_chunks")? else {
+            return Ok(None);
+        };
+        let count: usize = String::from_utf8_lossy(&count).parse().map_err(|_| {
+            IndexError::CorruptIndex(format!(
+                "search_cache_chunks is not a number: {:?}",
+                String::from_utf8_lossy(&count)
+            ))
+        })?;
+
+        // A missing chunk is an error, never a short cache. Deserializing a truncated
+        // stream would either fail somewhere confusing or, worse, succeed against a
+        // partial inverted index and drop search hits with nothing to show for it.
+        let mut serialized = Vec::new();
+        for i in 0..count {
+            let chunk = self.db.get(Self::search_cache_chunk_key(i))?.ok_or_else(|| {
+                IndexError::CorruptIndex(format!(
+                    "search cache chunk {i} of {count} is missing; the index is \
+                         incomplete and must be rebuilt"
+                ))
+            })?;
+            serialized.extend_from_slice(&chunk);
+        }
+        let cache: SearchCache = bincode::deserialize(&serialized)?;
+        Ok(Some(cache))
     }
 
     /// Load a single signature from RocksDB by its MD5 sum.
@@ -4195,6 +4276,93 @@ mod tests {
         // Test that we can save state without errors
         index.save_state()?;
 
+        Ok(())
+    }
+
+    /// A cache too big for one RocksDB value survives a write/read round trip.
+    ///
+    /// The real trigger is 4 GiB; this drives the same code path with a 1 KB chunk size so
+    /// the test costs milliseconds. What it is actually checking is that reassembly
+    /// preserves byte order across chunk boundaries -- a reversed or skipped chunk still
+    /// deserializes into *something* for many payloads, so a round trip that only asserts
+    /// "no error" would pass while returning a wrong index.
+    #[test]
+    fn test_search_cache_survives_chunking() -> Result<()> {
+        use tempfile::tempdir;
+        let dir = tempdir()?;
+        let index = ProteomeIndex::new(dir.path().join("chunked.db"), 10, 1, "protein20", false)?;
+
+        // Big enough to need several chunks, and varied enough that a misordered
+        // reassembly cannot coincidentally match.
+        let target_list: Vec<String> = (0..500).map(|i| format!("md5-{i:08x}")).collect();
+        let inverted_index: HashMap<u64, Vec<u32>> =
+            (0..500u64).map(|i| (i * 7919, vec![i as u32, (i as u32) + 1])).collect();
+        let kmer_frequencies: HashMap<u64, usize> =
+            (0..500u64).map(|i| (i * 7919, (i as usize) % 13 + 1)).collect();
+        let cache = crate::index::SearchCache {
+            target_list: target_list.clone(),
+            inverted_index: inverted_index.clone(),
+            kmer_frequencies: kmer_frequencies.clone(),
+        };
+        let serialized = bincode::serialize(&cache)?;
+        assert!(serialized.len() > 4096, "payload must span several 1 KB chunks");
+
+        index.write_search_cache_chunked(&serialized, 1024)?;
+        let loaded = index.load_search_cache()?.expect("a chunked cache must load");
+
+        assert_eq!(loaded.target_list, target_list);
+        assert_eq!(loaded.inverted_index, inverted_index);
+        assert_eq!(loaded.kmer_frequencies, kmer_frequencies);
+        Ok(())
+    }
+
+    /// A cache that fits keeps the single-key layout, so a database written by this build
+    /// is still readable by one without the chunking support.
+    #[test]
+    fn test_small_search_cache_stays_single_key() -> Result<()> {
+        use tempfile::tempdir;
+        let dir = tempdir()?;
+        let index = ProteomeIndex::new(dir.path().join("small.db"), 10, 1, "protein20", false)?;
+
+        let cache = crate::index::SearchCache {
+            target_list: vec!["md5-0".to_string()],
+            inverted_index: HashMap::from([(42u64, vec![0u32])]),
+            kmer_frequencies: HashMap::from([(42u64, 1usize)]),
+        };
+        let serialized = bincode::serialize(&cache)?;
+        index.write_search_cache_chunked(&serialized, 1 << 30)?;
+
+        assert!(index.db.get(b"search_cache")?.is_some(), "must use the original key");
+        assert!(index.db.get(b"search_cache_chunks")?.is_none(), "must not write a count");
+        assert!(index.load_search_cache()?.is_some());
+        Ok(())
+    }
+
+    /// A cache whose chunks are incomplete is an error, not a short index. Deserializing a
+    /// truncated stream can succeed against a partial inverted index and silently drop
+    /// search hits, which is far worse than refusing to open.
+    #[test]
+    fn test_missing_chunk_is_an_error_not_a_short_cache() -> Result<()> {
+        use tempfile::tempdir;
+        let dir = tempdir()?;
+        let index = ProteomeIndex::new(dir.path().join("torn.db"), 10, 1, "protein20", false)?;
+
+        let cache = crate::index::SearchCache {
+            target_list: (0..500).map(|i| format!("md5-{i:08x}")).collect(),
+            inverted_index: (0..500u64).map(|i| (i * 7919, vec![i as u32])).collect(),
+            kmer_frequencies: (0..500u64).map(|i| (i * 7919, 1usize)).collect(),
+        };
+        let serialized = bincode::serialize(&cache)?;
+        index.write_search_cache_chunked(&serialized, 1024)?;
+
+        index.db.delete(b"search_cache_chunk_1")?;
+        match index.load_search_cache() {
+            Ok(_) => panic!("a torn cache must not load"),
+            Err(err) => assert!(
+                err.to_string().contains("missing"),
+                "error should name the missing chunk, got: {err}"
+            ),
+        }
         Ok(())
     }
 }
--- a/src/rust/index.rs
+++ b/src/rust/index.rs
@@ -557,12 +557,34 @@ impl ProteomeIndex {
             chunk_size,
             chunks.len()
         );
+
+        // Order matters, and it is chosen so that an interrupted write leaves a database
+        // that REFUSES to open rather than one that opens wrong.
+        //
+        // The count goes first, as a delete. Rewriting a database that already had a valid
+        // count would otherwise leave that count standing over chunks being overwritten
+        // underneath it, and a crash midway would produce a count matched by a mixture of
+        // old and new chunks -- which deserializes into a plausible, wrong index.
+        self.db.delete(b"search_cache_chunks")?;
+        // Any stale single-key cache from an earlier build would otherwise win on read,
+        // because the reader prefers it.
+        self.db.delete(b"search_cache")?;
+
         for (i, chunk) in chunks.iter().enumerate() {
             self.db.put(Self::search_cache_chunk_key(i), chunk)?;
         }
-        // Any stale single-key cache from an earlier build of the same database would
-        // otherwise win on read, because the reader prefers it.
-        self.db.delete(b"search_cache")?;
+        // A previous write may have made more chunks than this one. Those are not read --
+        // the count bounds the loop -- but they are left behind claiming to be part of an
+        // index they no longer belong to, and the torn-write check below reads their
+        // presence as evidence. Clear them until the first key that is absent.
+        for i in chunks.len().. {
+            if self.db.get(Self::search_cache_chunk_key(i))?.is_none() {
+                break;
+            }
+            self.db.delete(Self::search_cache_chunk_key(i))?;
+        }
+
+        // Written LAST: its presence is what certifies every chunk above is on disk.
         self.db.put(b"search_cache_chunks", chunks.len().to_string().as_bytes())?;
         Ok(())
     }
@@ -1479,6 +1501,22 @@ impl ProteomeIndex {
         }
 
         let Some(count) = self.db.get(b"search_cache_chunks")? else {
+            // No count key. That is either a database written before the cache existed --
+            // legitimately Ok(None), and the caller falls back to loading every signature
+            // -- or one whose chunk write was interrupted before the count was committed.
+            //
+            // Those two must not be confused. A torn index returning Ok(None) is the
+            // quietest possible failure: the search still runs, on the slow path, against
+            // an index nobody is told is broken. Chunk 0 is written before any other, so
+            // its presence without a count means the write did not finish.
+            if self.db.get(Self::search_cache_chunk_key(0))?.is_some() {
+                return Err(IndexError::CorruptIndex(
+                    "search cache chunks are present but the chunk count is missing: the \
+                     index was interrupted while being written and is incomplete. Rebuild \
+                     it with `kmerseek index`."
+                        .to_string(),
+                ));
+            }
             return Ok(None);
         };
         let count: usize = String::from_utf8_lossy(&count).parse().map_err(|_| {
@@ -4335,6 +4373,90 @@ mod tests {
         assert!(index.db.get(b"search_cache")?.is_some(), "must use the original key");
         assert!(index.db.get(b"search_cache_chunks")?.is_none(), "must not write a count");
         assert!(index.load_search_cache()?.is_some());
+        Ok(())
+    }
+
+    /// An interrupted write must refuse to open, not read as a database with no cache.
+    ///
+    /// This is the quietest failure the chunked layout can produce: chunks on disk, no
+    /// count key, `Ok(None)` returned, and the caller falls back to loading every
+    /// signature. The search then runs -- slowly, and against an index nobody is told is
+    /// broken. A genuinely cache-free database must still return `Ok(None)`, so the two
+    /// cases are asserted together; a check that cannot tell them apart is no check.
+    #[test]
+    fn test_torn_write_is_loud_but_no_cache_is_quiet() -> Result<()> {
+        use tempfile::tempdir;
+        let dir = tempdir()?;
+
+        let torn = ProteomeIndex::new(dir.path().join("torn.db"), 10, 1, "protein20", false)?;
+        let cache = crate::index::SearchCache {
+            target_list: (0..500).map(|i| format!("md5-{i:08x}")).collect(),
+            inverted_index: (0..500u64).map(|i| (i * 7919, vec![i as u32])).collect(),
+            kmer_frequencies: (0..500u64).map(|i| (i * 7919, 1usize)).collect(),
+        };
+        let serialized = bincode::serialize(&cache)?;
+        torn.write_search_cache_chunked(&serialized, 1024)?;
+
+        // Exactly what a crash between the last chunk and the count leaves behind.
+        torn.db.delete(b"search_cache_chunks")?;
+        match torn.load_search_cache() {
+            Ok(None) => panic!("a torn write must not read as an absent cache"),
+            Ok(Some(_)) => panic!("a torn write must not load"),
+            Err(err) => assert!(
+                err.to_string().contains("interrupted"),
+                "error should say the write was interrupted, got: {err}"
+            ),
+        }
+
+        // A database that never had a cache is still legitimately empty.
+        let fresh = ProteomeIndex::new(dir.path().join("fresh.db"), 10, 1, "protein20", false)?;
+        assert!(
+            fresh.load_search_cache()?.is_none(),
+            "a database with no cache must return Ok(None), not an error"
+        );
+        Ok(())
+    }
+
+    /// Rewriting a database with a SMALLER cache must not leave the old cache's extra
+    /// chunks behind.
+    ///
+    /// They are never read, since the count bounds the loop, so the index is correct
+    /// either way -- but they are indistinguishable from the debris of an interrupted
+    /// write, which is what the torn-write check keys on. Left in place they would make a
+    /// healthy database fail that check the next time it was rewritten.
+    #[test]
+    fn test_rewriting_smaller_clears_stale_chunks() -> Result<()> {
+        use tempfile::tempdir;
+        let dir = tempdir()?;
+        let index = ProteomeIndex::new(dir.path().join("shrink.db"), 10, 1, "protein20", false)?;
+
+        let big = crate::index::SearchCache {
+            target_list: (0..2000).map(|i| format!("md5-{i:08x}")).collect(),
+            inverted_index: (0..2000u64).map(|i| (i * 7919, vec![i as u32])).collect(),
+            kmer_frequencies: (0..2000u64).map(|i| (i * 7919, 1usize)).collect(),
+        };
+        let big_bytes = bincode::serialize(&big)?;
+        index.write_search_cache_chunked(&big_bytes, 1024)?;
+        let big_chunks = big_bytes.len().div_ceil(1024);
+
+        let small = crate::index::SearchCache {
+            target_list: (0..100).map(|i| format!("md5-{i:08x}")).collect(),
+            inverted_index: (0..100u64).map(|i| (i * 7919, vec![i as u32])).collect(),
+            kmer_frequencies: (0..100u64).map(|i| (i * 7919, 1usize)).collect(),
+        };
+        let small_bytes = bincode::serialize(&small)?;
+        index.write_search_cache_chunked(&small_bytes, 1024)?;
+        let small_chunks = small_bytes.len().div_ceil(1024);
+        assert!(small_chunks < big_chunks, "second write must need fewer chunks");
+
+        for i in small_chunks..big_chunks {
+            assert!(
+                index.db.get(ProteomeIndex::search_cache_chunk_key(i))?.is_none(),
+                "stale chunk {i} from the larger cache was left behind"
+            );
+        }
+        let loaded = index.load_search_cache()?.expect("the smaller cache must load");
+        assert_eq!(loaded.target_list, small.target_list);
         Ok(())
     }
 
--- a/src/rust/errors.rs
+++ b/src/rust/errors.rs
@@ -23,6 +23,13 @@ pub enum IndexError {
     #[error("No saved state found in database")]
     NoSavedState,
 
+    /// The database opened, but what it holds is not a usable index. Distinct from
+    /// `Database`, which is RocksDB itself refusing, and from `NoSavedState`, which means
+    /// there is legitimately nothing stored: this is a cache that is present and broken,
+    /// and the only repair is a rebuild.
+    #[error("Corrupt index: {0}")]
+    CorruptIndex(String),
+
     #[error("IO error: {0}")]
     Io(#[from] std::io::Error),
 


src/rust/main.rs
```diff
@@ -799,6 +963,130 @@ fn assign_encoding(
...
+             A search will have to fit its own lambda and K (--ka-queries) or be given --ka-k.",
+            report.n_queries, report.n_regions
+        ),
+    }
+    Ok(())
+}
+
+/// One row per bin of x = lambda_pair S: the count of regions at or above it, the fitted
+/// line's count, the reference count, and whether the bin was inside the fit window.
+fn write_survival_csv(path: &std::path::Path, fit: &KaCalibration) -> IndexResult<()> {
```
olgabot: should this really be in main.rs? It's not the CLI

src/rust/main.rs
```diff
@@ -799,6 +963,130 @@ fn assign_encoding(
...
+                .map_or("none".to_string(), |b| format!("{:.1}", b as f64 * fit.bin_width))
+        );
+    }
+}
+
+/// Fit lambda and K on `n_queries` calibration queries of the index just built and store
+/// the fit in the index. Also prints the closed-form lambda and K for the database's own
+/// composition, so the effect of the seed requirement and of real sequence structure on
+/// each is visible.
+fn calibrate_index(
```
olgabot: Should this really be in main.rs? It's not called directly in the CLI

olgabot: Move to karlin_altschul.rs and maybe rename that to evalue.rs for clarity

src/rust/tests/test_cli.rs
```diff
@@ -539,3 +539,290 @@ fn test_cli_search_bcl2_ced9() -> Result<(), Box<dyn std::error::Error>> {
...
+
+    let exact = run(&[], &temp_dir.path().join("exact.csv"))?;
+    let extended = run(
+        &["--extend-mismatch-penalty", "2", "--extend-xdrop", "8"],
+        &temp_dir.path().join("extended.csv"),
+    )?;
+
+    // Off: the same 242 records (243 lines with the header) as test_cli_search_bcl2_ced9,
+    // all exact.
+    assert_eq!(exact.len(), 242);
```
olgabot: Can the 242 be a variable instead of. amagic number that we can use across multiple tests? Look for other re-used magic numbers

src/rust/tests/test_cli.rs
```diff
@@ -539,3 +539,290 @@ fn test_cli_search_bcl2_ced9() -> Result<(), Box<dyn std::error::Error>> {
...
+        assert!(r.region_length >= 12);
+        assert!(r.region_n_shared_kmers <= r.region_length - 12 + 1);
+        assert_eq!(r.region_subseq.len() as u32, r.region_length);
+        assert_eq!(r.target_subseq.len() as u32, r.region_length);
+    }
+    let bcl2_exact: Vec<_> =
+        exact.iter().filter(|r| r.target_name.contains("BCL2_HUMAN")).collect();
+    let bcl2_ext: Vec<_> =
+        extended.iter().filter(|r| r.target_name.contains("BCL2_HUMAN")).collect();
+    assert!(bcl2_ext.len() <= bcl2_exact.len());
```
olgabot: Show me the new BCL2/Ced9 region in this test

src/rust/tests/test_cli.rs
```diff
@@ -539,3 +539,290 @@ fn test_cli_search_bcl2_ced9() -> Result<(), Box<dyn std::error::Error>> {
...
+            "--alphabet",
+            "hp",
+            "--ka-queries",
+            "25",
+        ])
+        .assert()
+        .success()
+        .stderr(predicate::str::contains(
+            "Fitting Karlin-Altschul lambda and K on 25 database sequences (penalty 2, X-drop 8), \
+             censored against the same sequences shuffled-dipeptide",
```
olgabot: Clarify what "censored" means

olgabot: See the HTML explainer of evalues

src/rust/index.rs
```diff
@@ -510,12 +517,89 @@ impl ProteomeIndex {
...
+    /// Largest slice written under one RocksDB key.
+    ///
+    /// RocksDB refuses any single value at or above 4 GiB -- its length is a u32 -- and
+    /// returns `Invalid argument: value is too large`. That is a format limit, so no amount
+    /// of memory or tuning avoids it. A reviewed Swiss-Prot index (483_966 targets,
+    /// 115_749_594 unique k-mers) serializes to 4.43 GB and hit it.
+    ///
+    /// 1 GiB leaves a wide margin and keeps the chunk count small: even a 100 GB cache is
+    /// 100 keys.
+    const SEARCH_CACHE_CHUNK: usize = 1 << 30;
```
olgabot: I'm not sure about this chunk count. Why not have it be 2 or 4? Maybe that's. aseparate PR

src/rust/index.rs
```diff
@@ -510,12 +517,89 @@ impl ProteomeIndex {
...
+    /// count key, which reads as "no cache" rather than as a cache that silently ends
+    /// early -- the failure that would otherwise surface much later as a truncated
+    /// inverted index and quietly missing search hits.
+    fn write_search_cache_bytes(&self, serialized: &[u8]) -> IndexResult<()> {
+        self.write_search_cache_chunked(serialized, Self::SEARCH_CACHE_CHUNK)
+    }
+
+    /// The chunk size is a parameter so the split path can be tested against a few KB
+    /// instead of the 4 GiB it takes to reach it in production.
+    fn write_search_cache_chunked(&self, serialized: &[u8], chunk_size: usize) -> IndexResult<()> {
```
olgabot: What is. therelationship between  this and write_search_cache_bytes?

src/rust/index.rs
```diff
@@ -4193,6 +4362,269 @@ mod tests {
 
         Ok(())
     }
+
+    /// A cache too big for one RocksDB value survives a write/read round trip.
```
olgabot: Why are all these RocksDB changes. inthis PR?

src/rust/index.rs
```diff
@@ -4193,6 +4362,269 @@ mod tests {
...
+            bend_score: None,
+            rms_residual: 0.086,
+            survival: vec![(12, 9561)],
+            reference_survival: Vec::new(),
+            reference_lambda: None,
+            reference: None,
+        };
+        assert_eq!(index.ka_calibrations()?, Vec::new());
+        index.put_ka_calibration(&fit(2.0, 8.0, 0.0197))?;
+        index.put_ka_calibration(&fit(3.0, 8.0, 0.0673))?;
```
olgabot: This feelsl ike cheating. Why are you just putting these values in here?

src/rust/karlin_altschul.rs
```diff
@@ -0,0 +1,768 @@
...
+    let mean_score_tilted = a * lambda.exp() - penalty * b * (-penalty * lambda).exp();
+    Some(mean_score_tilted * (1.0 - (-lambda).exp()))
+}
+
+/// Which sequences are searched to fit lambda and K.
+///
+/// Measured on 300 human BCL2-related proteins at hp k=12, penalty 2, X-drop 8, 200
+/// calibration queries each (local slope of ln count vs score, per 4-score window):
+///
+/// - `Database`: 0.35 to 0.36 up to score 20, then the line bends upward into 1,800
```
olgabot: Reference figures in the docs/ikmages here

olgabot: Also. Ifeel like we should add some docs in general

src/rust/karlin_altschul.rs
```diff
@@ -0,0 +1,768 @@
...
+            DecoyNull::ShuffledDipeptide => write!(f, "shuffled-dipeptide"),
+        }
+    }
+}
+
+/// One fitted (lambda, K), stored in the index under `ka_calibration` and looked up at
+/// search time by (mismatch_penalty, xdrop). The seed length and alphabet are the index's.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct KaCalibration {
+    pub mismatch_penalty: f64,
```
olgabot: Suggest defaults for mismatch_penalty (2) and xdrop (8). Add doc comments

olgabot: Make. sureALL properties have a doc comment. Nothing should be empty

src/rust/karlin_altschul.rs
```diff
@@ -0,0 +1,768 @@
...
+    /// where the fit stops (`fit_scores_with_reference`). Empty and None for other nulls.
+    pub reference_survival: Vec<(i64, u64)>,
+    pub reference_lambda: Option<f64>,
+    /// How the reference queries were made (`Shuffled` or `ShuffledDipeptide`).
+    pub reference: Option<DecoyNull>,
+}
+
+impl KaCalibration {
+    /// The factor a search multiplies every pair's closed-form lambda by: the fitted slope.
+    pub fn lambda_scale(&self) -> f64 {
```
olgabot: didn't we rename this to `r_database`?

src/rust/evalue.rs
```diff
@@ -0,0 +1,768 @@
...
+/// The bins above the most populated one. Below the peak sit the bare seeds and the pairs
+/// whose lambda is small; the Karlin-Altschul tail is what comes after it.
+fn tail_bins(bins: &[(i64, u64)]) -> Vec<(i64, u64)> {
+    match bins.iter().enumerate().max_by_key(|(_, b)| b.1) {
+        Some((peak, _)) => bins[peak + 1..].to_vec(),
+        None => Vec::new(),
+    }
+}
+
+/// Least squares of ln count on score over `points`; returns (slope, intercept).
```
olgabot, line 308: State that the slope is K adn intercept is lambda

src/rust/evalue.rs
```diff
@@ -0,0 +1,768 @@
...
+/// A uniformly random rearrangement of `raw` with the same dipeptide counts, the same first
+/// residue and the same last residue (Kandel, Matias, Unger & Winkler 1996).
+///
+/// The residues are the vertices of a graph and each neighbouring pair an edge, so a
+/// rearrangement with the same dipeptide counts is a path that uses every edge once. Such a
+/// path exists when the edges that leave each vertex for the last time form a tree pointing
+/// at the final residue. That tree is drawn with Wilson's loop-erased random walk, weighted
+/// by edge multiplicity, which is the distribution the theorem needs; the other edges out
+/// of each vertex are then shuffled and the path is walked from the first residue.
+pub fn dipeptide_shuffle(raw: &str, rng: &mut SplitMix64) -> String {
```
olgabot, line 534: Do we really have to reinvent this wheel? Can't we juse use a library?

src/rust/karlin_altschul.rs
```diff
@@ -0,0 +1,768 @@
...
+    }
+    next
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_k_theory_matches_textbook_plus_minus_one_case() {
```
olgabot: Thes numbers seem okay but I really need to see the original citation to trust it. GIve the originals and no magic numbers!

olgabot: This applies to ALL the tests here

src/rust/karlin_altschul.rs
```diff
@@ -0,0 +1,768 @@
...
+        }
+        counts
+    }
+
+    /// Human BCL2 (P10415, tests/testdata/fasta/bcl2.fasta). Same length, same first and
+    /// last residue, every dipeptide as often as before, a different order, and the same
+    /// order again from the same seed.
+    #[test]
+    fn test_dipeptide_shuffle_keeps_every_dipeptide_count() {
+        let bcl2 = "MAHAGRTGYDNREIVMKYIHYKLSQRGYEWDAGDVGAAPPGAAPAPGIFSSQPGHTPHPAASRDPVARTSPLQTPAAPGAAAGPALSPVPPVVHLTLRQAGDDFSRRYRRDFAEMSSQLHLTPFTARGRFATVVEELFRDGVNWGRIVAFFEFGGVMCVESVNREMSPLVDNIALWMTEYLNRHLHTWIQDNGGWDAFVELYGPSMRPLFDFSWLSLKTLLSLALVGACITLGAYLGHK";
```
olgabot: Why are. youdefining this inline? Don't we havea BCL2 variable. youcane use?

src/rust/search.rs
```diff
@@ -62,6 +66,211 @@ impl Default for SearchFilters {
...
+/// outward from each seed along the stored encoded sequences, +1 per agreeing position and
+/// `-mismatch_penalty` per disagreeing one, and stops when the running score has fallen
+/// `xdrop` below its best (Altschul's X-drop). Two seeds on one diagonal whose extensions
+/// meet become one region.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct ExtensionParams {
+    /// Score subtracted per disagreeing encoded position. Must be positive.
+    pub mismatch_penalty: f64,
+    /// Extension stops once the running score is this far below its best so far.
+    pub xdrop: f64,
```
olgabot: Wait aren't these parameters on something else before? Why are they repeated?

src/rust/search.rs
```diff
@@ -62,6 +66,211 @@ impl Default for SearchFilters {
...
+    /// the `a` of the database against itself.
+    pub match_probability: f64,
+    pub n_queries: usize,
+    pub n_regions: usize,
+}
+
+/// What a calibration run is asked for; see `ProteinSearcher::calibrate_ka`.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct KaCalibrationSettings {
+    pub mismatch_penalty: f64,
```
olgabot: Why. aretehre so many repeated properties? can't htey be inherited or derived? This is hard to maintain

src/rust/search.rs
```diff
@@ -62,6 +66,211 @@ impl Default for SearchFilters {
...
+                    "; {} reference slope {:.3} over the same bins",
+                    c.reference.map_or("shuffled".to_string(), |r| r.to_string()),
+                    l / c.bin_width
+                ))
+            ),
+        }
+    }
+}
+
+/// Karlin-Altschul sum statistic for `r` segments whose normalised scores add to `t`:
```
olgabot: wtf. is r and wtf is t

src/rust/main.rs
olgabot: > [!NOTE]

Moved. `write_survival_csv` is now `evalue::write_survival_csv` (bf20748), next to the calibration it serialises; main.rs only passes `--ka-survival-out` through.

src/rust/main.rs
olgabot: > [!NOTE]

Moved with `warn_on_short_fit` and `write_survival_csv` into the library module (bf20748). main.rs now only builds `KaCalibrationSettings` from the flags and calls `evalue::calibrate_index`.

olgabot: > [!NOTE]

Done: `karlin_altschul.rs` is `evalue.rs` (bf20748), and the three functions live there. `karlin_altschul_lambda` stays in search.rs because the search solves it per pair; the module doc says so.

src/rust/tests/test_cli.rs
olgabot: > [!NOTE]

Done. `CED9_ROWS_HP_K12_MAX_PVALUE_0_7 = 242` at the top of test_cli.rs, with the note about 364 unfiltered rows that used to sit inline; `test_cli_search_bcl2_ced9` asserts `+ 1` for the header and the extension test asserts the constant. The k-mer size in that test is a local `KSIZE`. The other repeated numbers in the file are the fixture fit values (9838 regions, r_database 0.806, K 0.0115), which each appear once per CLI call they check; I left those as literals.

src/rust/tests/test_cli.rs
olgabot: > [!NOTE]

Now in the test (bf20748). The BH1 landmark of `test_cli_landmark_region_across_scaled`, CED9 162..181 against BCL2 138..157, is asserted before extension (19 residues, 8 shared k-mers, 0 mismatches) and after: it grows 7 residues to the right across two class flips, to 162..188 / 138..164, 26 residues, still 8 shared k-mers, 2 mismatches. The doc comment draws it:

```text
Ced9 pr: …QCPMSYGRLIGLISFGGFV AAKMMES VE…
Ced9 hp: …pphhphhphhhhhphhhhh hhphhpp hp
         ||||||||||||||||||| |.||.||
BCL2 hp: …pphhphhphhhhhphhhhh phpphpp pp
BCL2 pr: …RDGVNWGRIVAFFEFGGVM CVESVNR EM…
```

The 7 new positions score 5 − 2 × 2 = +1, so the walk keeps them.

src/rust/tests/test_cli.rs
olgabot: > [!NOTE]

"Censored" is gone. The message now reads: `Fitting r_database and K on 25 database sequences (mismatch penalty 2, give-up margin 8); the fit stops where their counts rise above the same sequences shuffled-dipeptide`. Same wording in the code comments and in `docs/evalue.md`.

olgabot: > [!NOTE]

Matched to the explainer: it says "where the fit stopped", "the stop rule", "the reference, the same queries shuffled", and calls X the give-up margin, so the CLI messages, doc comments and `docs/evalue.md` now use those words. `slope` is `r_database` everywhere too (struct field, survival CSV column, plot scripts, committed curves).

src/rust/index.rs
olgabot: > [!NOTE]

This chunking is not part of this PR any more. It was PR #48, which sat on this branch as a duplicate commit because the branch had not merged main; after merging main (b0cffb8) the diff no longer contains it. PR #51 then removed the search-cache blob altogether (schema 3 stores the inverted index in 1024 shards), so the chunk size question does not arise.

src/rust/index.rs
olgabot: > [!NOTE]

Same answer as the chunk-size comment: both functions were PR #48 showing up as a duplicate, and main has since replaced the whole writer in PR #51. After the merge with main this hunk is gone from the PR.

src/rust/index.rs
olgabot: > [!NOTE]

They were not meant to be. The branch was cut before #48 and #53 merged and carried its own copies of their commits, so GitHub showed them as part of this PR. Merging main (b0cffb8) removes them from the diff; what remains of this PR in index.rs is the `ka_calibration` key, the three methods that store and look up a fit, and one test.

src/rust/index.rs
olgabot: > [!NOTE]

Rewritten (bf20748). What that test checks is storage: a fit is filed under its (mismatch penalty, give-up margin), a second fit for the same pair replaces the first, a fit for another pair sits beside it, and a pair never fitted has none. The numbers in the record did not matter, but they looked like a real fit. The test now builds its records with `KaCalibration::placeholder(scoring, k)`, which carries zeros and empty curves and says so in its doc, and it is renamed `test_ka_calibration_is_keyed_by_penalty_and_give_up_margin`. The round trip of a real fit through the index is `search::tests::test_calibrate_ka_on_first25_is_stored_and_reused`, which fits on the 25-protein fixture and asserts every number.

src/rust/karlin_altschul.rs
olgabot: > [!NOTE]

Done: the `DecoyNull` doc points at `docs/images/ka_fit_grid_nulls_by_database.png` and `ka_fit_scope40_four_nulls.png` and gives the r_database each null produced on the three databases. The BCL2-fixture slopes that were there are replaced by the SCOPe40 / Swiss-Prot / UniRef50 numbers from the PR.

olgabot: > [!NOTE]

Added `docs/evalue.md` (f58d75f): every quantity in E = K m n e^(−λ_pair r_database S), where each comes from, the fit step by step, the choice of calibration queries, chaining, limits and the references, with the three figures. The README section links it and no longer says `--ka-k` defaults to 0.03. The module doc and `DecoyNull` point at it.

src/rust/karlin_altschul.rs
olgabot: > [!NOTE]

Done. `ExtensionScoring { mismatch_penalty, xdrop }` with `Default` = (2, 8) via `DEFAULT_MISMATCH_PENALTY` and `DEFAULT_XDROP` in search.rs; the `kmerseek index --extend-mismatch-penalty` and both `--extend-xdrop` flags read their defaults from those constants. `KaCalibration` stores one `scoring` instead of two loose fields, and every field has a doc comment.

olgabot: > [!NOTE]

Every field on `KaCalibration`, `ScoreFit`, `ExtensionParams`, `KaCalibrationSettings`, `KaParams`, `KaCalibrationReport` and `KaSource` has one now (bf20748).

src/rust/karlin_altschul.rs
olgabot: > [!NOTE]

Yes. The method was renamed in c66a69a and the field itself is now `r_database` too (was `slope`), along with the survival CSV column, the plot scripts and the committed curves under `docs/images/ka_fit_curves/`.

src/rust/evalue.rs
olgabot: > [!NOTE]

Stated on `line_through` now, the other way round from the comment: under the model ln(count at s) = ln(K L N (1 − e^−w)) − r_database · s, so minus the slope is r_database (the factor on every pair's λ), and the intercept, once L, N and the bin width are divided out, is K. `ScoreFit` and `KaCalibration` say the same on their fields.

src/rust/evalue.rs
olgabot: > [!NOTE]

I looked and found nothing to use. `cargo search` for ushuffle, dinucleotide shuffle, eulerian shuffle and dipeptide returns no crate that does the dipeptide-preserving shuffle (uShuffle is C and Python only). For the random numbers `rand` is already in the lockfile through statrs, but rand allows the stream for a given seed to change in minor releases, and a stored fit records only its seed, so "same seed, same fit" would break on a dependency bump. The 15-line splitmix64 plus the Wilson-tree shuffle is what keeps the fit reproducible; the doc on `dipeptide_shuffle` cites Kandel et al. 1996 for the algorithm.

src/rust/karlin_altschul.rs
olgabot: > [!NOTE]

Rewritten (bf20748). The ±1 test derives K from p and q: Karlin & Altschul 1990 give K = E[X e^(λX)] (1 − e^−λ) when the only positive step is +1, and with e^λ = q/p that is (q − p)(1 − p/q) = (q − p)²/q, the same closed form Ewens & Grant reach in their BLAST chapter; the assertion is `(q - p) * (q - p) / q` with p and q named. The a = 0.5, C = 2 test derives the golden ratio from y³ − 2y² + 1 = 0 in y = e^λ and computes the expected K from `a`, `penalty` and φ; 0.1631 stays as the check value with its derivation and the 4-million-step walk that confirmed it (458 segments seen, 478 predicted).

src/rust/karlin_altschul.rs
olgabot: > [!NOTE]

Done: the test reads BCL2_HUMAN from `TEST_BLC2_FASTA` (`tests/testdata/fasta/bcl2.fasta`) through a small `read_single_fasta` helper and checks it is 239 residues. There was no full-sequence constant to reuse; the two 30-residue fragments elsewhere are their own fixtures.

src/rust/search.rs

They were repeated, on `ExtensionParams`, `KaCalibrationSettings` and `KaCalibration`. Now (bf20748) there is one `ExtensionScoring { mismatch_penalty, xdrop }` and the three structs each carry a `scoring` field; a stored fit is looked up by `scoring` equality. K and r_database were also loose on `ExtensionParams`; it carries a `KaParams` now, and `KaCalibration::ka_params()` produces it.

src/rust/search.rs

Derived now. `KaCalibrationSettings` carries `scoring: ExtensionScoring` plus what only a calibration run needs (null, reference, n_queries, seed); `KaCalibration` is the settings plus the fit results; `ExtensionParams` is `scoring` plus `KaParams` plus the two chaining knobs. Rust structs cannot inherit from each other, so the shared part is its own struct that the others hold as a field; each number lives on one struct and every field is documented.

src/rust/search.rs

Renamed and explained. The signature is `karlin_altschul_sum_p(sum_normalised_score, n_segments)`: the chance that `n_segments` chained regions between an unrelated pair add up to at least `sum_normalised_score`, where each region's normalised score is λ S_i − ln(K m n), its raw score in nats less the size of the search. The doc gives the formula with r and t named right next to it, since that is how Karlin & Altschul 1993 write it.

src/rust/karlin_altschul.rs
olgabot: > [!NOTE]

Done for the rest of the module: the two synthetic fit tests name every number they make up (`N_SEED`, `SEED_SCORE`, `TRUE_LAMBDA`, `BUMP`, `K_RATIO`, `RAMP_START`, `RAMP_HEIGHT`, `RAMP_DECAY`) and assert against those names; `test_fit_scores_needs_enough_bins` uses `MIN_BIN_COUNT` and says it is short of `MIN_FIT_POINTS`; the least-squares slope in `test_fits_refuse_a_rising_curve` already had its derivation in the comment and keeps it.

PR #56, opened 2026-09-17
Expand B/J/Z when removing low-complexity k-mers

olgabot:
## Bug

With `--remove-low-complexity` on, `add_protein` hashed each window as-is. A window covering B, J or Z went into the sketch with the literal ambiguous byte. No query k-mer matches that hash, and the position map, which looks up each reading's hash, never recorded those windows. A search against such a target silently missed every window covering the ambiguous residue.

The other two per-window loops (the disambiguating branch used when removal is off, and the position map) did expand B/J/Z into their readings. Only the low-complexity branch skipped it.

## Fix

The three loops each had their own copy of the windowing and encoding. This replaces the copies with two helpers in `sketch.rs`, so every path walks the same readings:

- `hash_encoder`: the per-residue encoder (the old `residue_encoder` closure, with its capitalization rule).
- `Readings`: windows a sequence and calls back with `(position, encoded reading)`. A window with no ambiguous residue is encoded into one buffer per sequence; one with an ambiguous residue goes through `disambiguate_kmer`, which since #57 encodes first and returns one reading per resolution the alphabet tells apart. The whole-sequence ambiguity check is kept, so the common case pays nothing extra.

`add_protein` is now the branch dispatch plus the md5; the loops live in `add_windows`, `add_windows_without_low_complexity` and `record_positions`. `ProteinSketch::residue_encoder` and `hash_kmer` had no callers left.

The low-complexity rule is now one check on the encoded reading: dropped when it is a run of one class. A raw homopolymer always encodes to a run, so the separate raw check only mattered for dayhoff6, the one reduced alphabet the encoded check skipped: `EEEDD` encodes to `ccccc` there and was kept, while the same window under an HP alphabet was dropped. Every alphabet now behaves the same. README and `--help` say so.

Counters keep their meaning: `kmer_windows_examined` counts windows, `low_complexity_kmers_removed` counts readings dropped, both stay 0 with removal off.

## Test

`test_remove_low_complexity_keeps_both_readings_of_ambiguous_residues` in `index.rs` indexes `PLANTANDANIMALGENBMES` and `PLANTANDANIMALGENZMES` at k=5, scaled=1, under every alphabet in `Alphabet::all()`, with removal on and off, against a table of expected counts worked out by hand from the published partitions. No raw window is a homopolymer, so under protein20 removal drops nothing and both settings give 21 k-mers: 17 windows, four of which cover the ambiguous residue and contribute two readings each. Alphabets that put D and N in one class (dayhoff6, gbmr7, mmseqs12) give 17; the HP alphabets merge more windows and removal drops the ones that encode to a one-class run (one under the Lehninger splits, four under gbmr4).

Against the code before this branch, every alphabet comes out at 13 or fewer with removal on: only the windows not covering the ambiguous residue got position entries.

`test_remove_low_complexity_checks_encoded_runs_under_dayhoff` in `sketch.rs` indexes the FKBP8 poly-glutamate tract under dayhoff6 and expects 9 of 26 windows dropped (the 13-residue run of `c`); the old code drops 7.

## Checks

- `cargo fmt --check`: clean
- `cargo clippy --no-default-features --all-targets`: no warnings
- `cargo test --no-default-features --lib -- --test-threads=2`: 268 passed


src/rust/index.rs
```diff
@@ -2362,6 +2400,35 @@ mod tests {
         Ok(())
     }
 
+    /// With removal on, a window covering an ambiguous residue is still indexed under both
+    /// readings. Hashing the literal B or Z instead put a hash in the sketch that no query
+    /// k-mer matches and that the position map, which looks up each reading's hash, never
+    /// recorded, so a search silently missed every window covering the residue.
+    #[test]
+    fn test_remove_low_complexity_keeps_both_readings_of_ambiguous_residues() -> Result<()> {
```
olgabot, line 2239: add tests of all other alphabets here, using the same bitscore size as protein20 and k=5

olgabot: > [!NOTE]

Done in dfc0b70. The test now runs over `Alphabet::all()` at k=5, scaled=1, against a table of expected counts per alphabet, worked out by hand from the published partitions rather than read off kmerseek.

The counts differ by alphabet:
- protein20 and the finer alphabets (wwmj5, funcgroups8, sdm12, wass14, hsdm17, uniprot18): 21. The two readings of B or Z are different k-mers.
- dayhoff6, gbmr7, mmseqs12: 17. D and N (and E and Q) share a class, so both readings collapse to one k-mer. gbmr7 gives 16 on the Z sequence because `ANIMA` and the E reading of `ENZME` both encode to `adaaa`.
- The HP alphabets: 14 or 15, since many unrelated windows encode alike. Removal then drops one reading (`IMALG` → `hhhhh`) under the Lehninger splits and four under gbmr4.

Against the code before this branch, every alphabet comes out at 13 or fewer with removal on (13 for protein20, 9 for gbmr4).

Writing the table showed that the `is_hp_moltype` field was misnamed: it is true for every table-backed alphabet, not only the HP family, which is why gbmr4 drops four readings. Renamed to `drops_encoded_runs` and fixed its doc. dayhoff6 is the one reduced alphabet without the encoded-run check, which was already so before this branch; left as is.

Also merged main twice to clear the conflicts (04af206, then 1ced0ab after #57 landed). #57 made `disambiguate_kmer` encode first and return readings ready to hash, so `Readings::for_each` now hands its callback the window as written and one encoded reading, and `KmerEncoder` shrinks to the residue encoder plus the low-complexity check. `ProteinSketch::residue_encoder` and `hash_kmer` had no callers left and are gone. The expected table did not change: a reading that collapses under the alphabet hashed the same either way.

PR #57, opened 2026-09-17
Disambiguate after encoding, and test expansion with a 22-residue-ambiguous ribonuclease under every alphabet

olgabot:
## Summary

Three things, driven by one test protein.

1. A search test for a protein dense in ambiguous residues, under all 19 alphabets.
2. Disambiguation now happens after alphabet encoding, and the ceiling on ambiguous residues per k-mer goes from 10 to 20.
3. Matched regions run through ambiguous residues instead of stopping at them.

Topi pancreatic ribonuclease (UniProt P00659, RNAS1_DAMKO) was sequenced by Edman degradation and 22 of its 124 residues came out as B (Asp or Asn) or Z (Glu or Gln): 12 B and 10 Z, never more than 17 residues apart. The test indexes the 125 UniProtKB ribonucleases it was downloaded with (`tests/testdata/fasta/ribonuclease_125_entries_uniprotkb_2026_09_17.fasta.gz`) and searches with P00659 (`tests/testdata/fasta/rnas1_damko_P00659.fasta`).

Each alphabet runs at the k where a k-mer carries as many bits as protein20 at k=10 (k·log2(N) = 43.2 bits, rounded): k=43 for the two-class HP alphabets, 27 for hp_lehninger_hpc3, 22 for the four-class ones, 19 for wwmj5, 17 for dayhoff6, 15 for gbmr7, 14 for funcgroups8, 12 for sdm12 and mmseqs12, 11 for wass14 and hsdm17, 10 for uniprot18 and protein20.

## Disambiguate after encoding

`disambiguate_kmer` now takes the alphabet's per-residue encoder and returns encoded readings. An ambiguous residue whose two readings encode to the same class is written as that class and does not count toward the ceiling. Under every HP alphabet, dayhoff6, gbmr4, gbmr7 and mmseqs12, B and Z are not ambiguous at all. Before this change, P00659 at k=43 under hp_lehninger2 hashed 28,032 readings for 82 windows (all hashing to 71 values) and dropped the 11 windows holding 11 or 12 ambiguous residues. Now it hashes 82 readings and drops nothing.

`MAX_AMBIGUOUS_RESIDUES_PER_KMER` goes from 10 to 20, counting residues still ambiguous after encoding. P00659's densest window holds 12 at k=43, the most in Swiss-Prot.

## Regions through ambiguous residues

Under any reduced alphabet, a hit involving an ambiguous residue got no matched region at all. The stored encoded sequence wrote B as `B` (or as `X` through sourmash's encoder), the target's read as a class, and both region paths (`find_matched_regions` and `exact_run_around`) compared encoded residues for equality. So the expansion raised containment but never produced a region except under protein20, which stores no encoded sequence.

Now the stored sequence writes an ambiguous residue as its class where the alphabet merges the two readings (`p` under HP) and as the letter itself where it does not (`B` under sdm12), and both region paths let the letter agree with either class it stands for, through one shared encoder (`hash_functions::residue_encoder`) so the check stays precise.

Searching P00659 against the ribonucleases, with the filters off:

| alphabet | k | target | regions (residues, 1-based) | why the breaks |
|---|---|---|---|---|
| protein20 | 10 | goat RNAS1_CAPHI | 1–102, 104–124 | residue 103 is K in topi, E in goat |
| protein20 | 10 | bovine RNAS1_BOVIN | 4–18, 20–36, 38–102, 104–124 | four real mismatches: 3, 19, 37, 103 |
| sdm12 | 12 | goat | 1–124 | sdm12 merges K and E |
| sdm12 | 12 | bovine | 1–18, 20–36, 38–102, 104–124 | sdm12 merges S and T at residue 3 |
| hp_lehninger2 | 43 | bovine | 20–124 | S/A at residue 19 is the only mismatch HP can see |

Before this PR every reduced-alphabet row here was empty.

## What the test pins, per alphabet

Goat RNase (P67926) is the same 124 residues as topi, agrees at every ambiguous position, and differs at one real position (103). Bovine RNase A (P61823) has a signal peptide, then a mature chain that differs at four real positions (3, 19, 37, 103) and agrees at every ambiguous one. So shared k-mer counts and region spans are computable by hand.

For every alphabet the test asserts the readings sketched, the distinct query hashes and the k-mers shared with bovine (for example 82 under protein20, 63 under HP, 103 of 103 under gbmr4). Two more tests pin goat's 105 shared k-mers under protein20 and the region spans in the table above. Every expected value was recomputed independently in Python from the two sequences and the alphabet tables in `alphabets.rs`.

## Also fixes the test build on main

PR #51 changed `ProteinSearcher::new` to return `IndexResult`, and `test_prefix_sums_add_every_reading_of_an_ambiguous_residue` (merged from the region-tfidf branch) still called it without `?`, so `cargo test --lib` did not compile on main. One-character fix in the first commit.

## Test plan

- `cargo test --no-default-features --lib -- --test-threads=2`: 269 passed
- `cargo clippy --no-default-features --all-targets`: clean
- `cargo fmt --check`: clean


src/rust/tests/test_ambiguous_residues.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    /// A matched region runs through an ambiguous residue instead of stopping at it. Every
+    /// region below covers several B and Z, and each break is a real mismatch the alphabet
+    /// can see. Bovine differs from topi at residues 3 (S/T), 19 (S/A), 37 (Q/K) and 103
+    /// (K/N): sdm12 merges S with T and so breaks at the other three, giving four regions;
+    /// every HP alphabet merges all but S/A, giving one region from residue 20 to the end.
+    /// Goat differs only at residue 103 (K/E), which sdm12 merges, so under sdm12 goat is
+    /// one region covering the whole protein.
+    #[test]
+    fn test_matched_regions_run_through_ambiguous_residues() -> Result<()> {
```
olgabot, line 286: how does the expansion affect the pvalue, region enrichment, etc?

olgabot: > [!NOTE]

This PR changes no formula. I built `origin/main` and this branch and ran both on the 125 ribonucleases; under protein20 every column of every row is identical between the two. What the PR changes is which readings exist (one per window where the alphabet merges the pair) and whether a reduced-alphabet hit shows up at all: a hit with no matched region writes no CSV row, so on `main` topi's hits under sdm12, dayhoff6 and every HP alphabet were missing from the output, not only their regions.

## What expansion does to each column

Goat RNase is topi with every B and Z resolved, and it has the same four mismatches against bovine (residues 3, 19, 37, 103), so goat→bovine is the no-ambiguity control. Both hits share 82 k-mers and the same four regions (residues 4–18, 20–36, 38–102, 104–124) with the same `region_n_shared_kmers`. protein20, k=10, N=125 targets, filters off:

| column | goat→bovine (0 ambiguous) | topi→bovine (22 ambiguous, 541 readings) |
|---|---|---|
| n_intersecting_hashes | 82 | 82 |
| containment | 0.713 | 0.152 |
| jaccard | 0.471 | 0.137 |
| max_containment | 0.713 | 0.582 |
| query_expected_shared_kmers | 3.39 | 6.96 |
| query_enrichment | 24.2 | 11.8 |
| query_tfidf | 415 | 2454 |
| region 38–102: expected | 1.98 | 4.26 |
| region 38–102: enrichment | 28.4 | 13.2 |
| region 38–102: tfidf | 188 | 1512 |
| region 38–102: mean_idf | 3.37 | 27.0 |
| region 4–18: poisson_score | 7.49 | 4.68 |
| region 104–124: poisson_score | 13.35 | 10.92 |

sdm12 at k=12 shows the same pattern: containment 0.68 → 0.11, region 38–102 enrichment 28 → 10, mean_idf 3.3 → 39.

Why each one moves:

- **containment, jaccard.** The denominator is the query's hash count, which is readings (541), not windows (115). A true homolog matches one reading per window, so topi's containment can never exceed 115/541 = 0.21. This does not depend on database size. `max_containment` partly recovers through target-in-query.
- **query_expected_shared_kmers, query_enrichment, query_poisson_pvalue.** Expected = Σ over query hashes of max(freq, 1)/N, so every reading adds at least 1/N. Topi's 426 extra readings add 3.4 at N=125. At Swiss-Prot's N (575,748) the floor adds 0.0007; only readings that occur in other proteins count. So this dilution scales as extra readings / N and matters for small databases only. (Topi is itself one of the 125 targets here, so every reading has freq ≥ 1 either way.)
- **region expected, enrichment, poisson_score.** Same sum, restricted to k-mer positions inside the region: `build_prefix` adds every reading at a position (PR #52; that keeps the whole-query total equal to the sum above). A k=10 window holding 4 ambiguous residues carries 16 readings, so its floor is 16/N instead of 1/N. Shrinks with N the same way.
- **region_tfidf, region_mean_idf.** This one does not shrink with N. IDF is ln(N / number of targets holding the k-mer), so a k-mer seen in one target scores ln(125) = 4.8 here. `tfidf` sums IDF over every reading at each position that exists in the database; `mean_idf` divides that by `n_shared`, which counts positions. So a region through ambiguous residues looks rarer than it is, by up to the number of readings per window: 3.4 → 27 here. With the query in the database every reading exists. Without it, only readings present in other proteins count, but each of those occurs in few proteins and so has a high IDF, and the direction is the same.

## Merging alphabets

Under HP, dayhoff6, gbmr4, gbmr7 and mmseqs12, B and Z are not ambiguous after this PR, so there is nothing to dilute: topi→bovine under hp_lehninger2 equals goat→bovine column for column (containment 0.768, expected 1.23, enrichment 51.3, one region 20–124, score 307.7, mean_idf 4.08).

Before this PR the same alphabets hashed every reading to the same value and pushed it 2ⁿ times. Topi's stored sketch under dayhoff6 had 108 hashes but 1,452 positions and an abundance sum of 1,452; under hp_lehninger2, 71 hashes and 28,032. That inflated `average_abund` for any hit involving topi, and would have multiplied the region's expected count and IDF at those positions by 2ⁿ had a region ever formed. Now every hash carries one position and abundance 1.

(One side effect in this test set: under HP, topi and goat now encode to the same sequence, share an md5, and the index keeps only the first. Topi is absent from the HP index on this branch. That is why goat→bovine under HP differs slightly between `main` and this branch: goat's k-mers lost one occurrence each.)

## Follow-ups, not in this PR

1. Count windows, not readings, in containment's denominator (and jaccard's union). That puts topi→bovine at 82/115 = 0.713, the same as goat.
2. Decide what `region_mean_idf` should mean for a position with several readings: divide by readings, or take the matched reading's IDF only. Today it divides reading-summed IDF by position count.

Happy to open an issue for these if you want them tracked separately.

olgabot: > [!NOTE]

Follow-up 1 is done in #76: containment, its target-side counterpart and jaccard count a window once however many readings it holds, on both sides. Topi→bovine is 82/115 = 0.713, the same as goat, and topi against a copy of itself is 1.

Follow-up 2, `region_mean_idf`, drawn out for the same region (38–102, protein20 k=10). Box height is the IDF of one reading; the filled box is the k-mer bovine holds, the hollow boxes are the other readings of that window. Every height is recomputed from the FASTA by `scripts/plot_region_mean_idf_readings.py` (in #76) and the sums match this CSV row: 1512.11 for topi, 188.47 for goat.

region_mean_idf: readings stacked per window, topi vs bovine

What the picture shows: the filled boxes are all 3.0–3.7, because every ribonuclease in the set shares those k-mers. The hollow boxes are all 4.8, the maximum, because only topi holds them. A window with four ambiguous residues carries 16 readings and towers over its neighbours for no biological reason.

The three ways to average, and what each says:

1. Every reading ÷ windows, today: 27.0. An ambiguous region looks eight times rarer than the same region in goat. This measures how many B and Z the sequencer left, not how rare the k-mers are.
2. Matched reading only ÷ windows: 3.37, identical to goat. "How rare are the k-mers this hit is built from", which is what the column name says.
3. Every reading ÷ readings: 4.55. An average over k-mers the hit never used; bounded, but it still moves with the number of ambiguous residues.

I would go with 2. `region_tfidf` becomes the IDF sum over the matched readings and `region_mean_idf = region_tfidf / n_shared` keeps its definition. The cost: `idf_prefix`, the per-query running total built once in `prepare_query`, cannot answer it, because which reading matched depends on the target. The sum has to come from the region's own shared hashes, `n_shared` lookups per region, small next to finding the region. `region_expected_shared_kmers` would keep summing every reading, since any reading can match by chance, so the two running totals part ways on purpose.

Say the word and I will open that as its own PR on top of #76.

PR #58, opened 2026-09-17
Add kmerseek pair, visualize_pair.py and a Foldseek-style search report

olgabot:
## What

Per-pair and per-search views of the k-mers kmerseek matches on.

**`kmerseek pair`** compares one query sequence with one target sequence at any alphabet and k-mer size and writes JSON: both sequences with their encodings, every shared k-mer with its 0-based position in each, and the matched regions from `find_matched_regions` (longest first). The encoding tables stay in Rust; the Python side reads the encoding off the JSON.

**`scripts/visualize_pair.py`** draws that JSON (PNG, SVG, and with `--html` a self-contained page):

- A dot plot, query position against target position. With `--domains`, each protein is a line with boxes for its Pfam domains along its axis and each domain's span is shaded across the plot, so a run sits in a named cell ("Bcl-2 × Bcl-2") without reading coordinates. Runs of two or more consecutive shared k-mers are numbered diagonal segments; lone shared k-mers are dots.
- One alignment block per run, longest first, numbered to match: query row, identical residues written between the rows, target row, 1-based coordinates at both ends, residue boxes coloured hydrophobic or polar. The header gives the two regions, the length, identical residues and polar residues (1 of 14 polar is an all-hydrophobic run).

**`scripts/visualize_search.py`** turns a search CSV into one HTML report per query on the query axis: the query drawn once at the top with its domains, under a histogram of how many database entries have a run over each residue (light any run, dark a run with ≥5 identical residues; the noise map that discounts the BCL-2 loop). One row per protein (entries of one gene fold together) with the numbers in the row (length, runs, longest run, identical residues in it, shared k-mers, ranking statistic) and every run as a bar at its query coordinates, shaded by its share of identical residues, stacked in lanes where runs overlap. Click a row and the pair view opens beneath it. Rows sort by `region_evalue` when the CSV has it (after #54) and by the BH q-value until then; the column headers also sort by identical/length/runs/shared, which put p53 and POU4F1 among the family members and show why the ranking statistic is the default. The page is `scripts/kmerseek_hits_template.html` with a title token and a data token filled by the script; it filters rows by name, identical residues, the statistic, Swiss-Prot status, fragments and query domain, has a dark theme and keyboard keys, and downloads the rows and runs as TSV, the runs as FASTA, the overview as SVG or PNG and the data as JSON. The same template renders a single pair (`{"pair": model}`) as a one-row page with the row open, so `visualize_pair.py --html` goes through it too and `visualize_pair_html.py` is gone; `hits_page.py` holds the one render call. The template reads score columns from the data (`R.score_columns`, `row.scores`) and falls back to the single q-value, which is what the writer sends; new scores come in their own PRs. An open row also shows both full sequences, coloured by class with each run underlined in the shade of its bar; a run hovered in the dot plot, its alignment block or the sequences lights up in all three, and the dot plot answers the pointer anywhere with a crosshair and the nearest run, single k-mer or USalign residue pair. The red line in the dot plot is USalign's residue pairs from superposing the two AlphaFold models (solid where USalign marks the pair close), and the PNG, model and README call it that. Example, CED-9 against the 25 BCL-2-like proteins in `tests/testdata/fasta` at k=15 (15 rows; BCL-2's BH1 run is 8th by q-value, behind longer polar-rich runs such as RTN3's 23 residues with 2 identical): [docs/examples/ced9_kmerseek_hits_example.html](https://htmlpreview.github.io/?https://github.com/seanome/kmerseek/blob/olgabot/pair-visualization/docs/examples/ced9_kmerseek_hits_example.html).

`--domains` takes Pfam-style tables (TSV, CSV or parquet) with a protein column (`accession`/`protein`/`name`), `domain_start`/`domain_end` or `start`/`end` (1-based inclusive) and a `name`/`pfam_name`/`pfam_id` column. Proteins match by full header, first token or UniProt accession, so the `*_pfam_domains.parquet` tables in 2024-kmerseek-analysis (built from Pfam-A.regions) work unchanged.

Example, human BCL-2 against C. elegans CED-9 at `hp_lehninger2`, k=12, both proteins' Pfam domains. Run 1 is the BH1 motif inside the Bcl-2 domain of both: 5/19 residues identical, 19/19 the same hydrophobic/polar class.

BCL2_HUMAN vs CED9_CAEEL

```bash
kmerseek pair -q tests/testdata/fasta/bcl2.fasta -t tests/testdata/fasta/ced9.fasta --alphabet hp --ksize 12 -o bcl2_vs_ced9.json
python scripts/visualize_pair.py --pair bcl2_vs_ced9.json --output-dir pair_png/ --domains scripts/testdata/bcl2_ced9_pfam_domains.tsv --html
kmerseek index -i tests/testdata/fasta/bcl2_first25_uniprotkb_accession_O43236_OR_accession_2025_02_06.fasta.gz --alphabet hp --ksize 15 -o bcl2_25.rocksdb
kmerseek search -q tests/testdata/fasta/ced9.fasta -t bcl2_25.rocksdb -o results.csv --alphabet hp --ksize 15
python scripts/visualize_search.py --csv results.csv --query-fasta tests/testdata/fasta/ced9.fasta --target-fasta tests/testdata/fasta/bcl2_first25_uniprotkb_accession_O43236_OR_accession_2025_02_06.fasta.gz --output-dir report/ --domains scripts/testdata/bcl2_25_pfam_domains.tsv scripts/testdata/bcl2_ced9_pfam_domains.tsv
```

## Notes

- `find_matched_regions` reports every lone shared k-mer as a region exactly k residues long, and the JSON keeps them so it matches `search` output. The figures draw those as singles and give alignments only to runs of two or more consecutive k-mers.
- The two encoded rows and the same-class ticks from the first version are gone: inside a run every column is the same class by definition of a shared k-mer. With `--flank N` the middle line becomes BLAST's (`:` where only the class agrees), so the run reads as a solid middle line and the flank as a ragged one.
- Region names come from the domain with the most overlap; a run outside every domain is "no region".
- `serde_json` moves from dev-dependencies to dependencies.
- `kmerseek pair` JSON carries the alphabet's whole class table (`classes`), so the legends name every residue of each class, not only the ones the two sequences contain.
- Example pair page: [docs/examples/bcl2_vs_ced9_pair_example.html](https://htmlpreview.github.io/?https://github.com/seanome/kmerseek/blob/olgabot/pair-visualization/docs/examples/bcl2_vs_ced9_pair_example.html) (rendered through htmlpreview.github.io).
- The first commit (760e727) fixes `cargo test` on `main`: PR #52's test `ambiguous_residue_readings_all_count_toward_expected_shared_kmers` calls `ProteinSearcher::new(index)` without `?`, and PR #51 made it return `IndexResult`. One character; easy to cherry-pick if you want it separate.

## Tests

- 7 unit tests in `pair.rs` and 2 CLI tests: exact BH1 coordinates (BCL-2 138–157, CED-9 162–181, 0-based half-open), 27 shared hp 12-mers in 14 regions, the APG repeat cross-product at protein20 k=3, named-record lookup.
- 20 pytest tests for the figure model, matplotlib figure and HTML (`scripts/test_visualize_pair.py`): domain table reading and matching, region naming, run headers, the flank middle line, `</script>` in a name cannot close the embedded script or the page title.
- 4 tests for the USalign wrapper (on a checked-in USalign report).
- 8 pytest tests for the search report (`scripts/test_visualize_search.py`) on a checked-in 362-row CSV (CED-9 against the 25-protein BCL-2 fixture): header parsing, gene folding, ranking fallback, the per-residue histogram, and one end-to-end render that skips without a `kmerseek` build.
- 272 Rust tests pass, clippy clean, `cargo fmt` run.
- `scripts/test_visualize_hits.py::test_fixture_matches_rust_csv_schema` fails on `main` as well; fixed in #59.

## Identities and structure

- Identities are counted on each run's own diagonal, with no gapped alignment. MCL-1's BH1 run counts 1 identical residue with BCL-2 though NWGR is in both rows, because the exact hp run sits 3 residues off the true alignment; the README says so, so a low count reads as "the run is not the alignment", not "unrelated proteins". The report's identical column and bar shading use this count.
- `--structures DIR` runs USalign/TM-align on the pair's AlphaFold or PDB files (`scripts/structure_alignment.py`) and draws every aligned residue pair as a thin red path across the dot plot; each run's header says whether it lies on the path. BCL-2 vs CED-9: run 1 (BH1) is on it, run 3 is 4 residues off, which settles which of the two is real. The report gets a TM-score column.

## Fixed in their own PRs

- #62: `find_matched_regions` split runs on repeated k-mers (BCL-2 against itself: 144 aa, not 239). Once it merges, this PR's `kmerseek pair` fixture needs regenerating.
- #63: identical sequences collapsed to one name in the index, and a FASTA query's own entry was skipped as a self-match.
- #61: the one-character test fix `main` needs to compile its tests (this PR's first commit is the same change).


scripts/gapped_alignment.py
```diff
@@ -0,0 +1,162 @@
+"""Gapped alignment of two short protein stretches, for counting identities.
```
olgabot: What do we really need this for?

olgabot: Aren't we only doing ungapped alignments? Why is BLOSUM here? Why are we reinventing the wheel? Can't we use a library so we don'thave to maintain thsi?

scripts/gapped_alignment.py
```diff
@@ -0,0 +1,162 @@
...
+NEG = -(10**6)
+
+
+def _codes(seq):
+    """Residue indices into BLOSUM62; anything unknown (U, O, lower case) scores as X."""
+    return np.array([_INDEX.get(c.upper(), _INDEX["X"]) for c in seq], dtype=np.intp)
+
+
+def _fill(a, b, local):
+    """Gotoh matrices: M (match ends here), X (gap in `a`, i.e. a run of `b` residues against
```
olgabot: What is "Gotoh"?

scripts/pair_model.py
```diff
@@ -0,0 +1,354 @@
...
+import statistics
+from collections import defaultdict
+
+import polars as pl
+
+from gapped_alignment import global_align
+from visualize_hits import short_label
+
+# What a biologist calls the classes of the hydrophobic/polar alphabets.
+CLASS_NAMES = {"h": "hydrophobic", "p": "polar", "c": "cysteine"}
```
olgabot, line 17: Always show the residues that are deteremined to be hydrophobic and polar for each alphabet, since it changes.

scripts/visualize_pair_html.py
```diff
@@ -0,0 +1,310 @@
+"""Self-contained interactive HTML for one pair, and the pieces the search report reuses.
```
olgabot: I want to see an example output HTML too

olgabot (commented):
did a cursory review, will do more. indepth later

scripts/gapped_alignment.py
olgabot: > [!NOTE]

What it is for: the search report's "identical residues" column and its solid/hollow bars. kmerseek itself only ever finds ungapped runs, and on a run's own diagonal MCL-1's BH1 counts 1 identical residue with BCL-2 even though NWGR is in both, because the exact hp run sits 3 residues off the true alignment. Allowing one gap in the run plus 10 residues either side gives 12. The count is only a label on the figure; it changes nothing in search or in the ranking. `--no-gapped` turns it off.

The aligner needs a substitution matrix to decide where the gap goes. BLOSUM62 with gap open 11 and extend 1 is BLAST's protein default.

Agreed on not maintaining it. 42fe74d replaces the 162-line file with Biopython's `PairwiseAligner` (same matrix and gap costs, end gaps charged). It gives byte-identical output on every existing test and the example PNG, so the wrapper is now 45 lines that map Biopython's result onto the dict `pair_model.py` reads. `biopython` is added to the three environment files.

scripts/gapped_alignment.py
olgabot: > [!NOTE]

Gotoh (1982) is the standard way to align when opening a gap costs more (11) than extending it by one residue (1); the field calls these affine gap costs. It keeps three score tables instead of one: the best score with the two residues paired, with a gap in the first sequence, and with a gap in the second. That comment described the hand-written tables, which are gone as of 42fe74d; Biopython does it now and the docstring says only which matrix and gap costs are used.

scripts/pair_model.py
olgabot: > [!NOTE]

Done in e81e331. The legend already listed the residues of each class, but it read them off the two sequences, so a pair missing a residue would leave it out of the legend (BCL-2 and CED-9 happen to contain all 20, which hid this). `kmerseek pair` now writes the alphabet's whole table as a `classes` field, found by encoding the 20 standard residues through the same path as the sequences, so it is right for every alphabet including `dayhoff6`, whose table lives in sourmash. The PNG legend, the pair page and the search report all draw from that field; for the example, "hydrophobic (A F G I L M P V W Y)" and "polar (C D E H K N Q R S T)".

scripts/visualize_pair_html.py
olgabot: > [!NOTE]

Checked in as [docs/examples/bcl2_vs_ced9_pair_example.html](https://github.com/seanome/kmerseek/blob/olgabot/pair-visualization/docs/examples/bcl2_vs_ced9_pair_example.html) (d265fdd), same inputs as the PNG: BCL2_HUMAN against CED9_CAEEL at hp k=12, both proteins' Pfam domains, the USalign path. GitHub serves the raw file as text, so to see it rendered: [open through htmlpreview.github.io](https://htmlpreview.github.io/?https://github.com/seanome/kmerseek/blob/olgabot/pair-visualization/docs/examples/bcl2_vs_ced9_pair_example.html). Hover a grey dot for its k-mer; click a run's number to jump to its alignment. The README links the same page.

scripts/gapped_alignment.py
olgabot: Oh hmm biopython is kind of huge and heavy. can you weigh the pros and cons of having such a big dependency for this one piece of functionality? Can we use it for something else, like viewing motifs or aligntments?

olgabot: > [!NOTE]

Measured before your "ungapped only" call came in, for the record: the biopython wheel is 2.6 MB and installs to 19 MB, its only dependency is numpy, and `import Bio.Align` takes 55 ms. Next to what the scripts already load (polars 130 MB, scipy 65 MB, numpy 32 MB, matplotlib 28 MB) it is not the heavy one, and `Bio.Align` was the only module in use. The other things it could have done here (`Bio.SeqIO` for the FASTA reading, `Bio.PDB` for the CIF files) are already covered by screed and USalign's own parsing, so there was no second use.

Moot now: 3f0b005 drops the gapped alignment altogether, per your decision. Identities are counted on the run's own diagonal only, `--gap-flank` and `--no-gapped` are gone from both scripts, `scripts/gapped_alignment.py` and its tests are deleted, and biopython is out of the environment files. The README keeps the MCL-1 BH1 example as a caution (1 identical on the diagonal though NWGR is in both rows), so a low count reads as "the run is not the alignment", not "unrelated proteins". Example PNG, SVG and HTML regenerated.

PR #62, opened 2026-09-18
Chain matched regions along each diagonal

olgabot:
`find_matched_regions` sorted the shared (query, target) position pairs by query position and walked them requiring (q+1, t+1). A k-mer that occurs twice in the target puts (q, t2) between (q, t1) and (q+1, t1+1), the consecutive check failed, and one exact match came out as several regions. BCL-2 against its own sequence at hp k=12 reported regions of 45 and 144 residues instead of one of 239; the Ala/Pro/Gly loop repeats several hp 12-mers.

The fix sorts by diagonal (target position minus query position) first, then query position, so every comparison in the walk stays on one diagonal. This is what the scaled path (`find_matched_regions_scaled`, PR #50) already did.

## What changes in output

- Regions grow and their count falls wherever a sequence repeats a k-mer. CED-9 against the 25-protein fixture: 330 regions instead of 362; with `--max-pvalue 0.7`, 218 rows instead of 242.
- Whole-query statistics (shared k-mers, containment, TF-IDF, query p-value) are unchanged.
- Region scores move with the merged lengths. `test_query_scope_alone_keeps_a_diffuse_match_with_no_standout_region` used BCL2A1 vs ASPP2, whose merged best region (13 aa, 5 k-mers) now scores p = 0.049, just over the 0.05 cap; it now uses BCL-2 vs RTN3 (152 shared 9-mers, 217 regions, best region p = 0.053), which still shows the query scope passing while no region does.

## Tests

- New: BCL-2 against itself is one 239-residue region plus 16 off-diagonal loop repeats; the BCL-2/CED-9 pair (80, 253)–(81, 254) stays one 13-residue run although (81, 170) sorts between them by query position.
- 265 pass with `SDKROOT=$(xcrun --show-sdk-path) cargo test --no-default-features --lib -- --test-threads=2`; clippy clean.

Stacked on #61 (the one-character test fix `main` needs to compile its tests). Once this merges, the `kmerseek pair` fixtures in #58 need regenerating (27 shared 12-mers stay, 14 regions become 13 and 5 runs become 6).


src/rust/search.rs
```diff
@@ -3599,11 +3644,15 @@ mod tests {
 
     /// Complements `test_pvalue_scopes_combine_with_or`'s BCL2/CED9 example, where the region
     /// scope rescues a hit the query scope rejects, with a real case running the other
-    /// direction. BCL2A1 vs ASPP2/TP53BP2 at k=9, in the same 25-sequence fixture database, is
-    /// an overwhelming whole-protein match (115 shared k-mers scattered across 333 short
+    /// direction. BCL2 vs RTN3 (reticulon-3) at k=9, in the same 25-sequence fixture database,
```
olgabot, line 3647: I want to see this example, show me the figure

olgabot: > [!NOTE]

Figure in 0f66daf, made with the new `scripts/plot_matched_regions.py` from a `kmerseek search` of the 25-protein fixture against itself at hp k=9, once with this PR's build and once with the parent commit's.

Matched regions of BCL2 vs RTN3 and BCL2A1 vs ASPP2, before and after chaining

Left column: every matched region as a segment at its query and target positions. Right column: each region's p-value, grey for this PR, orange outline for the parent commit, blue for the best region, dashed line at the default cutoff p = 0.05.

- BCL2 vs RTN3 (top): 152 shared k-mers, 217 regions (402 before chaining). Best region is 11 aa at p = 0.0533, still right of the cutoff, so only the whole-query scope (p = 1.0e-9) keeps this hit.
- BCL2A1 vs ASPP2 (bottom): 115 shared k-mers, 183 regions (333 before). Chaining merged its best region to 13 aa at p = 0.0491, just left of the cutoff, which is why the test switched pairs.

Chaining moved the whole histogram left: regions that a repeated k-mer used to split now count their k-mers together. Across the fixture this raised the number of reported pairs from 242 to 380, all through the region scope; whole-query p-values are unchanged.

PR #63, opened 2026-09-18
Report every entry that shares a sketch

olgabot:
The index stores one signature per sketch md5 (a wrapping sum of the sketch's mins), so a second entry with the same k-mer set was dropped and its name lost. A hit on that sketch was reported under the first name only.

## Changes

- `ingest` writes each skipped name as its own `alias_{md5}_{name}` key with an empty value. A sequence seen n times costs n small writes, nothing is rewritten, and no alias map is held during indexing or loaded when the index is reopened. `ProteomeIndex::aliases_of(md5)` reads a sketch's names with one prefix seek.
- `search_one` still returns one result per sketch. The CSV writer calls `ProteinSearcher::csv_rows`, which repeats each region row under every name of the target's sketch, and, in all-vs-all (where each query is an index entry), under every name of the query's sketch too. Same statistics on every repeat, since the k-mer sets are identical; only the names differ.
- `compare` always drops a target whose md5 equals the query's, in every search mode. A FASTA query identical to an index entry does not list that entry.

Not done here: when a separate query index is searched against a target index, only the target side's names are expanded. The query side's duplicate names live in the query index.

## Tests

- `test_repeated_sequence_is_indexed_once` checks the alias; new `test_aliases_persist_and_accumulate_across_reopen` checks a third copy joins the list after a reopen.
- New `identical_entries_share_one_result_and_every_csv_row`: CED-9 against BCL-2, a copy of BCL-2 and CED-9 gives one result whose CSV rows repeat under both BCL-2 names with equal statistics; a BCL-2 query gets only CED-9; all-vs-all gives two results and repeats the BCL-2-query rows under both query names.
- 265 pass with `SDKROOT=$(xcrun --show-sdk-path) cargo test --no-default-features --lib -- --test-threads=2`; fmt and clippy clean.

Stacked on #61. Independent of #62; the two touch different lines of `search.rs`.


src/rust/index.rs
```diff
@@ -1483,6 +1492,20 @@ impl ProteomeIndex {
         Ok(true)
     }
 
+    /// Every name stored as a duplicate of another sketch, by the md5 it duplicates: the
```
olgabot: why are we duplicating things?

src/rust/search.rs
```diff
@@ -38,6 +38,10 @@ pub struct SearchFilters {
     /// heuristic cutoff on a ranking score, not a statistically calibrated significance
     /// threshold.
     pub min_region_score: f64,
+    /// Drop a target whose sketch is the query's own (same md5). On for all-vs-all searches
+    /// of one index against itself, where every query would otherwise hit itself; off when a
+    /// FASTA is searched against an index, where the query's identical entry is a real hit.
+    pub skip_self_matches: bool,
```
olgabot: No always skip self-matches. taht's dumb and increases the row count by N

src/rust/search.rs
```diff
@@ -946,12 +984,11 @@ impl ProteinSearcher {
...
-        // Skip self-matches by comparing MD5 sums
-        // WHY: In all-vs-all searches, we don't want to compare a signature against itself.
-        // MD5 sum is a unique identifier for each signature, so comparing MD5 sums is the
-        // most reliable way to detect self-matches. This is idiomatic Rust - we use early
-        // returns to avoid unnecessary computation when we know the result will be invalid.
-        if query.sketch.signature().md5sum == target.signature().md5sum {
+        // In an all-vs-all search every query is also a target, and its hit on itself says
+        // nothing. When a FASTA is searched against an index, the query's identical entry is
+        // a real hit (BCL-2 against a database holding BCL2_HUMAN), so the skip is opt-in.
+        if filters.skip_self_matches && query.sketch.signature().md5sum == target.signature().md5sum
```
olgabot: skip self matches!!

src/rust/index.rs
olgabot: > [!NOTE]

Changed in edbe40b. Each later name is now its own `alias_{md5}_{name}` key with an empty value: one small write per repeated sequence, nothing rewritten, no alias map held during indexing or loaded at search open. `aliases_of(md5)` reads a sketch's names with one prefix seek, only for sketches that were hit. Results are no longer cloned per name; the CSV writer repeats the region rows under each name (`ProteinSearcher::csv_rows`).

src/rust/search.rs

Done in edbe40b. The flag is gone; `compare` always drops a target with the query's md5, in every search mode. A FASTA query identical to an entry no longer lists that entry.

src/rust/search.rs

Done in edbe40b, always skipped now (see the reply on `skip_self_matches` above).

PR #67, opened 2026-09-21
Extend matched regions past their exact seed with an X-drop walk

olgabot:
A matched region is a maximal exact run in the encoded alphabet, so one
class flip ends it. Between remote homologs the HP pattern agrees per
position far better than any 23-residue stretch agrees exactly, so an
exact run is a seed, not the match. `--extend-mismatch-penalty C` walks
outward from each seed along the stored encoded sequences, +1 per
agreeing position and -C per disagreeing one, and stops once the running
score has fallen `--extend-xdrop` (default 8) below its best, BLAST's
X-drop rule. Seeds on one diagonal whose extensions meet become one
region.

`MatchedRegion` and the CSV gain `n_mismatches`; `n_shared` keeps counting
exact shared k-mers, summed on merge and never recomputed from the grown
span, so the region Poisson test is unchanged for exact regions and only
more conservative for extended ones. Default off: without the flag every
region is exact and the new column reads 0.

First of a stack that splits PR #54. Scoring the extended region (bits,
E-value) comes in the next PR; this one only moves region boundaries.

## Where this sits

Stack that splits #54 into reviewable pieces. Each PR is based on the one before it; this one is on #77. After the PR before it merges, this one gets retargeted to `main`.

**1. Extend regions past the seed** (#67)
2. Bits and E-value with a closed-form lambda (#68)
3. Fit a line to a score survival curve (#69)

- https://github.com/seanome/kmerseek/pull/54#discussion_r4057719139

- Built on #77, so the four `n_mismatches: 0` initializers collapse to one line in `MatchedRegion::unscored`.
- `--extend-xdrop` below 0 is refused with a message (it would end every extension at its first mismatch), with a CLI test.


src/rust/tests/test_cli.rs
```diff
@@ -764,60 +764,136 @@ fn test_cli_index_rejects_bad_scaled() -> Result<(), Box<dyn std::error::Error>>
     Ok(())
 }
 
+/// `--extend-mismatch-penalty` grows regions past their exact seeds and reports the
+/// mismatches inside; without it every region is exact and the new column reads 0.
+///
+/// The BH1 match between CED9 and BCL2 (the landmark of
+/// `test_cli_landmark_region_across_scaled`, here in the `hp` alphabet) is an exact
+/// 19-residue run, CED9 162..181 against BCL2 138..157. With a mismatch penalty of 2 and a
+/// give-up margin of 8 it grows 7 residues to the right, across two class flips, and stops
```
olgabot, line 831: Why dosn't this grow to the left? Why only on the right? I need to see a figure

olgabot: > [!NOTE]
> Written by Claude Code (Fable 5.1) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

It does walk left. `extend_regions` calls `xdrop_walk` once per side, and both walks start at score 0 on the seed edge. The left walk just never finds anything worth keeping: the two residues before the seed (CED9 D, T against BCL2 F, L) are polar against hydrophobic, so the score starts at -4 and never climbs back above 0. A side keeps residues only up to its best running score, and the best on the left is the starting 0, so it keeps none. The walk then gives up eight residues out, ten below its best. On the right the score dips to -2, climbs to +1 after 7 residues, and the walk stops when it falls 9 below that peak, keeping the 7.

The X-drop walk on both sides of the BH1 seed. Top: the residues and their hydrophobic (h) or polar (p) class, with | for the same class and x for different. Bottom: the running score of each walk, the give-up line at best-so-far minus 8, the exact seed in grey, and the residues the walk adds in teal.

Pushed as c76c273: the doc comment here now shows both walks with their running scores, `xdrop_walk` says what a best score of 0 means, and `scripts/plot_xdrop_walk.py` draws the figure from the `kmerseek pair` JSON fixture, with tests that pin both score traces.

olgabot: Rename to docs/images/xdrop_walk_bcl2_ced9_bh1.png‎

olgabot: Also reference this figure in some doc somwhere. it hsouldn't be some bare file

olgabot: > [!NOTE]
> Written by Claude Code (Fable 5.1) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

Done in e6363e8. The figure is now `docs/images/xdrop_walk_bcl2_ced9_bh1.png` (and `.svg`), and the README's section "Extending matched regions past the exact seed" shows it right after the paragraph that defines the walk, with the left and right outcomes explained beside it. The script, the test's doc comment, and the note on `xdrop_walk` point at the new name.

The walk on both sides of the BH1 seed

PR #68, opened 2026-09-21
Score extended regions with Karlin-Altschul bits and an E-value

olgabot:
Extension alone moves region boundaries but not ranks: the Poisson test
still counts only exact shared k-mers, so a region that grew across
mismatches scores the same as before it grew. Each extended region now
also gets the score BLAST uses for an ungapped alignment, in the encoded
alphabet: S = matches - C x mismatches, E = K m n e^(-lambda S), and
bits = (lambda S - ln K) / ln 2, as `region_ka_bits` and `region_evalue`.

lambda is solved per pair from the two sequences' class compositions
(Schaffer et al. 2001), as the positive root of
a e^x + (1 - a) e^(-C x) = 1 with a the chance that a random position
from each falls in the same class. When a >= C / (1 + C) there is no
positive root: agreement is what the two compositions do by default, so
the region gets 0 bits and no significance. That is what makes this the
ranking statistic for extended regions rather than the Poisson count,
which sees a transmembrane helix against any other as a long exact run.

K depends on the alphabet, seed length, penalty, give-up margin and
database, so this PR takes it from `--ka-k`, required with
`--extend-mismatch-penalty`. Fitting K on the index itself is the next
PRs of the stack that splits PR #54.

The CLI test asserts the BH1 landmark's bits (18.7454) and E-value
(5.31363) under K = 0.03, with the arithmetic from the two compositions
in its doc comment.

## Where this sits

Stack that splits #54 into reviewable pieces. Each PR is based on the one before it; this one is on #67. After the PR before it merges, this one gets retargeted to `main`.

1. Extend regions past the seed (#67)
**2. Bits and E-value with a closed-form lambda** (#68)
3. Fit a line to a score survival curve (#69)

- https://github.com/seanome/kmerseek/pull/54#discussion_r4057721006
- https://github.com/seanome/kmerseek/pull/54#discussion_r4057721058

- `class_composition` returns a `[f64; 256]` on the stack instead of a `HashMap`, and the query's composition is computed once in `PreparedQuery` rather than once per candidate pair.
- `--ka-k` of 0 or less is refused (it silently produced no bits and infinite E-values), with a CLI test.


src/rust/main.rs
```diff
@@ -144,6 +144,14 @@ enum Commands {
         #[arg(long, default_value_t = DEFAULT_XDROP)]
         extend_xdrop: f64,
 
+        /// Karlin-Altschul K for `region_evalue` and `region_ka_bits` on extended regions,
+        /// with the closed-form lambda solved per pair from the two sequences' class
+        /// compositions. K depends on the alphabet, seed length, penalty, give-up margin and
+        /// database, so it has to be measured on decoys for the index in use. Required with
+        /// --extend-mismatch-penalty, and unused without it.
+        #[arg(long)]
+        ka_k: Option<f64>,
```
olgabot: Why is this required? It should just be used directly without having to specify, by readin gthe index, if you want extended alignemnts, which should be the default

src/rust/search.rs
```diff
@@ -107,6 +109,78 @@ impl Default for ExtensionScoring {
 pub struct ExtensionParams {
     /// Mismatch penalty and give-up margin.
     pub scoring: ExtensionScoring,
+    /// The constant of the E-value; from `--ka-k`.
+    pub ka: KaParams,
+}
+
+/// The constant of the E-value a search runs with: E = K m n e^(-lambda_pair S).
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct KaParams {
```
olgabot: This seems stupid to have a single value in a struct? Shouldn't like r_database and m and n be here too?

src/rust/search.rs
```diff
@@ -107,6 +109,78 @@ impl Default for ExtensionScoring {
...
+/// The Karlin-Altschul lambda for +1 / -penalty scoring when a random pair of positions
+/// matches with probability `a`: the positive root of a e^x + (1-a) e^(-penalty x) = 1.
+///
+/// A positive root exists only when the expected score a - penalty (1-a) is negative,
+/// i.e. a < penalty / (1 + penalty). Above that, agreement is what these two compositions
+/// do by default and no run of it is surprising: returns 0. The left side is convex with
+/// value 1 at x = 0 and slope a - penalty (1-a) there, so bisection on [0, hi] with hi
+/// pushed out until f(hi) > 1 is safe; e^hi overflows by hi = 1024, so with a > 0 that
+/// takes at most ten doublings.
+pub fn karlin_altschul_lambda(a: f64, penalty: f64) -> f64 {
```
olgabot: Rename from a -> u as per the explainer https://github.com/seanome/kmerseek/pull/84

src/rust/search.rs
```diff
@@ -107,6 +109,78 @@ impl Default for ExtensionScoring {
...
+    p.iter().zip(q).map(|(a, b)| a * b).sum()
+}
+
+/// The Karlin-Altschul lambda for +1 / -penalty scoring when a random pair of positions
+/// matches with probability `a`: the positive root of a e^x + (1-a) e^(-penalty x) = 1.
+///
+/// A positive root exists only when the expected score a - penalty (1-a) is negative,
+/// i.e. a < penalty / (1 + penalty). Above that, agreement is what these two compositions
+/// do by default and no run of it is surprising: returns 0. The left side is convex with
+/// value 1 at x = 0 and slope a - penalty (1-a) there, so bisection on [0, hi] with hi
```
olgabot: what is hi?

src/rust/search.rs
```diff
@@ -1148,6 +1254,32 @@ impl ProteinSearcher {
...
 
+        // Karlin-Altschul bits and E-value per region, on the pair's own class compositions.
+        // Only meaningful with a mismatch penalty, which is the score's mismatch term.
+        if let Some(params) = self.extension {
+            if let (Some(q_enc), Some(t_enc), Some(q_composition)) = (
+                query.sketch.get_moltype_sequence(),
+                target.get_moltype_sequence(),
+                query.composition.as_ref(),
+            ) {
+                let a = match_probability(q_composition, &class_composition(t_enc.as_bytes()));
```
olgabot:
```suggestion
                let match_prob = match_probability(q_composition, &class_composition(t_enc.as_bytes()));
```

src/rust/search.rs
olgabot:
```suggestion
pub fn karlin_altschul_lambda(match_prob: f64, penalty: f64) -> f64 {
```

actually let's call it match_prob, and mention it's called u in the explainer: https://seanome.github.io/kmerseek/karlin_altschul_explainer.html

README.md
```diff
@@ -167,6 +167,31 @@ disagree. The region Poisson score keeps counting exact k-mers against the expec
...
+
+```
+S    = matches - C × mismatches
+E    = K × m × n × e^(-λS)
+bits = (λS - ln K) / ln 2
+```
+
+`m` is the query length and `n` the number of k-mers in the database. `λ` is solved
+per pair from the two sequences' class compositions (Schäffer et al. 2001): the chance
+`a` that a random position from each falls in the same class gives λ as the positive
```
olgabot:
```suggestion
`u` that a random position from each falls in the same class gives λ as the positive
```

Rename as per the explainer

src/rust/main.rs
olgabot: > [!NOTE]
> Written by Claude Code (Opus 5.5) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

It is required only in this PR, because nothing else can supply K yet. Later PRs in the stack remove the requirement:

- #71 fits K (and `r_database`) on the target index before the search when `--ka-k` is not given, so the flag becomes optional.
- #72 fits K when the index is built and stores it in the index. Search reads it from there, so `--ka-k` is only an override.

Making extension the default belongs after #72, once every new index carries its own K. Before that, an index built without a fit would have to be fitted at every search. I'd add it as one more PR on top of #75: `--extend-mismatch-penalty` defaults to 2 when the index holds a fit, and `0` turns extension off.

src/rust/search.rs
> Written by Claude Code (Opus 5.5) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

`r_database` joins it in #71 (`KaParams { k, r_database }`). `m` and `n` are not settings of a search: `m` is each query's length and `n` is read from the index (`db_n_kmers`), so they stay arguments.

To give the struct a job in this PR, 427785a moves the two formulas that use K onto it: `KaParams::bits(lambda, S)` and `KaParams::evalue(lambda, S, m, n)`. `compare()` now calls those instead of writing the arithmetic out.

src/rust/search.rs
> Written by Claude Code (Opus 5.5) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

Done in 427785a with your second suggestion: the argument is `match_prob`, and the doc comment says the explainer calls it `u` and links to it. The test's comments and the CLI test's worked arithmetic use `u`.

src/rust/search.rs
> Written by Claude Code (Opus 5.5) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

The upper end of the range the bisection searches for the root. It is now `upper` (and `lo` is `lower`), and the doc comment says how it is found: start at 1 and double until the left side of the equation is above 1, which puts the root between 0 and `upper`.

src/rust/search.rs
> Written by Claude Code (Opus 5.5) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

Applied in 427785a (rustfmt splits it over two lines).

README.md
olgabot: > [!NOTE]
> Written by Claude Code (Opus 5.5) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

Applied in 427785a, along with the two other `a`s in that paragraph: `u e^λ + (1 - u) e^(-Cλ) = 1` and `u ≥ C / (1 + C)`.

PR #69, opened 2026-09-21
Fit a line to the tail of a score survival curve, stopping below the related pairs

olgabot:
Under Karlin-Altschul statistics the number of regions scoring exactly S
between unrelated sequences falls as e^(-lambda S), so ln(count) against S
is a straight line whose slope gives lambda and whose height gives K
(Altschul & Gish 1996; Pearson 1998). Related pairs lift the counts at
high S. `fit_scores` reads the line off the highest bins below that lift:
it grows the fit one bin at a time and stops at the first bin whose ln
count sits more than two units of counting noise above the line so far.
`fit_scores_with_reference` does the same on the ratio of two curves,
the real one and the same queries shuffled, because relatives below 40%
identity come in as a ramp the single-curve step test does not see.

Bins with fewer than MIN_BIN_COUNT (30) regions and everything up to the
most populated bin are left out: below the peak sit the bare seeds and
the pairs whose lambda is small. The line is read off the top FIT_WINDOW
(8) accepted bins, nearest the scores that decide a hit, and needs at
least MIN_FIT_POINTS (4). A rising or flat curve gives None.

Pure functions on already-binned scores, with tests on made-up counts
that follow the model exactly, every number named. The searcher that
produces the scores and the record that stores the fit are the next PRs
of the stack that splits PR #54.

## Where this sits

Stack that splits #54 into reviewable pieces. Each PR is based on the one before it; this one is on #68. After the PR before it merges, this one gets retargeted to `main`.

1. Extend regions past the seed (#67)
2. Bits and E-value with a closed-form lambda (#68)
**3. Fit a line to a score survival curve** (#69)

- https://github.com/seanome/kmerseek/pull/54#discussion_r4057719729
- https://github.com/seanome/kmerseek/pull/54#discussion_r4057722334

- `survival_counts` drops non-finite scores instead of letting `as i64` saturate and open a bin at `i64::MAX`, with a test.


src/rust/evalue.rs
```diff
@@ -0,0 +1,425 @@
...
+    let slope = if sxx > 0.0 { sxy / sxx } else { 0.0 };
+    (slope, my - slope * mx)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_survival_counts() {
```
olgabot: I'd really love figures for all these tests. It's ahrd to trust just pure text

olgabot: > [!NOTE]
> Written by Claude Code (Opus 5.5) at olgabot's request. Posted from her account, so replies here are to Claude, not to her.

Added in 63a255f: one figure per test, drawn from what the Rust code returned. When `KMERSEEK_FIGURE_DATA` is set, each test writes the counts it fits and the fit it got to JSON, and `scripts/plot_fit_scores_tests.py` draws them into `docs/images/fit_scores_tests/`. Each test's doc comment names its figure. With the variable unset the tests write nothing.

In every panel, filled teal circles are the bins the line was fitted on, hollow circles are bins it did not use, the orange triangle is the bin where the fit stopped, and the dotted line is the 30-region floor.

`test_fit_scores_recovers_slope_and_stops_at_homolog_excess`: the fit reads lambda 0.400 off bins 14-21 with or without 400 extra regions at score 27.
fit_scores

`test_fit_scores_with_reference_stops_where_the_ratio_rises`: on the right, the ratio to the shuffled counts. Related pairs from score 24 push bin 26 past its limit (grey dash), so the fit stops there and reads 0.385. Without them the ratio stays flat and the fit runs to bin 33 and reads 0.399.
fit_scores_with_reference

`test_fit_scores_needs_enough_bins`: three usable bins, and a fit needs four.
needs_enough_bins

`test_fits_refuse_a_rising_curve`: rising counts give no fit, and neither does rising against a flat reference. The same counts falling give lambda ln(10)/10 = 0.230.
rising

`test_survival_counts` and `test_bin_counts`: 15.5 falls into bin 15, and infinite or undefined scores are dropped.
counts

